PR #15, opened 2015-01-05
Ensure the permissions of a new redis-sentinel.conf

cdent:
Without these changes a from-scratch redis-sentinel.conf can be
created that is not writable by the redis user, as such the sentinel
server will exit immediately after starting.

This change ensures the permissions of the *.puppet file and then
preserves those permissions upon copy. Unfortunately the copied-to
file is hard to test in the spec files, so verification in test is
no longer available. It works properly in manual testing.

### DIFF
--- a/manifests/sentinel.pp
+++ b/manifests/sentinel.pp
@@ -136,16 +136,14 @@ class redis::sentinel (
   file {
     $config_file_orig:
       ensure  => present,
-      content => template($conf_template);
-
-    $config_file:
       owner => $service_user,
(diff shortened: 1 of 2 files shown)


cdent:
Sorry I didn't realize this in the original pull request @arioch. I guess all my testing was on existing installations. The bug will rarely show up since the distros install a default file in the first place, but in case it has been removed, this is useful to have (it bit me during packstack testing).

arioch:
Not to worry. :+1:

PR #16, opened 2015-01-07
Make sentinel maninfest ensure the redis package

cdent:
The sentinel service needs to make sure that the redis package has
been installed on the machine where sentinel is being run. All my
previous testing had assumed that there would be a redis server and
sentinel on the same machines. This doesn't have to be the case, so
the manifest needs to make allowances for that.

### DIFF
--- a/manifests/sentinel.pp
+++ b/manifests/sentinel.pp
@@ -150,6 +150,10 @@ class redis::sentinel (
       refreshonly => true;
   }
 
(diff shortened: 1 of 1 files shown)


cdent:
@arioch One more! This was found when testing with the multiple sentinels on hosts separate from the slaves.

Thanks.

arioch:
Keep 'em coming. :+1:

PR #17, opened 2015-01-09
Make config file owner dependency on redis package explicit

cdent:
Without this it's possible that on a clean machine where only sentinel
is being installed the install will fail because the 'redis' user
(who owns the *.puppet config file) will not yet exist.

By ensuring that the redis package is installed explicitly first the
user will be there.

The previous pull request related to this problem added the package
dependency but did not make the ordering explicity.


cdent:
One more @arioch! I hope this is the last one (at least for a while).

PR #20, opened 2015-01-15
Avoid resource dependency when having both Redis Server & Sentinel

EmilienM:
When running Redis Server & Redis Sentinel on the same node, avoid a
resource dupplication with package resource.


EmilienM:
wow, fast merge :-)
probably another patchset later

arioch:
keep 'em coming ;)

Issue #18, opened 2015-01-14
Sentinel init/upstart file doesn't exist


antonlindstrom:
When trying to install sentinel (`class { 'redis::sentinel': }`) on Ubuntu 14.04, the following error appears:

    Error: /Stage[main]/Redis::Sentinel/Service[redis-sentinel]: Could not find init script or upstart conf file for 'redis-sentinel'

I would expect that there would be an init or upstart script in place that would start the sentinel service.

Building a template for the sentinel service should probably do the trick.

Thank you for an awesome module!

arioch:
@cdent, could you please look into this?

cdent:
Yeah, can do tomorrow. Up til now all my testing has been rpm-based distros. I'll take a look at Ubuntu tomorrow.

BTW: There should be yet another pull request on the sentinel support coming from one of my colleagues tomorrow: The way the sentinel.pp requires the redis package is flaky and not properly puppet idiomatic.

Clearly I have no idea what I'm doing ... :)

antonlindstrom:
That sounds great, thanks!

cdent:
In the long term I reckon this is a bug in debian, so I've [reported one there](http://bugs.debian.org/cgi-bin/bugreport.cgi?bug=775414).

In the short term I guess we need to stub in an init script of our own. What's the best way to do that?

arioch:
@cdent, this should get you going pretty fast: https://github.com/jordansissel/pleaserun

cdent:
Thanks. I'm going to pause on implementing this until #19 has resolved (more comments over there in a moment) as the place to hook in the changes will be more clear then.

antonlindstrom:
Awesome, thanks!

PR #23, opened 2015-01-16
Add an initscript for sentinel on apt-based systems

cdent:
Until an initscript is available for redis-sentinel in its package[1]
we need to roll our own. This change adds that by creating a templated
initscript with the sentinel config file and service owner and group
as template variables that get filled in.

Testing this revealed some bugs and misfeatures that are also addressed
in this change:

* apt systems want the redis and redis-sentinel processes to daemonize
  (systemd-based) rpm does not. This was not reflected in the params
  defaults, nor reflected in the sentinel configuration template. Now
  it is.

* the package_name for redis-sentinel was not being passed in as a
  parameter, instead it was just using the params.pp default for
  redis server. When [1] is resolve, it will likely mean a different
  package for redis-sentinel so we need there to be a separate parameter
  for the sentinel package that gets installed. For now the name of
  the existing package is used.

[1] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=775414

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -14,7 +14,6 @@ class redis::params {
   $auto_aof_rewrite_percentage = 100
   $bind                        = '127.0.0.1'
   $conf_template               = 'redis/redis.conf.erb'
-  $daemonize                   = true
   $databases                   = 16
   $dbfilename                  = 'dump.rdb'
   $extra_config_file           = undef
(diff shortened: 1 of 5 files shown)


cdent:
@antonlindstrom and @arioch as far as my testing goes on ubuntu 14.04 this seems to fix #18 as expected. It would probably be wise for other folk besides me to test it before merging it though.

As you'll see the changes cascaded a bit to requiring a few other fixes.

cdent:
Please don't merge this yet, I left out a critical part: The initscript is not getting linked into the rc directories so the service won't launch on boot. I'll get that early tomorrow.

cdent:
The second commit should fix the the lack of initscript, and also fixes pid file handling. I reckon this is ready for someone else to give it a try and if it passes muster, should be good to merge.

antonlindstrom:
Looks, good, I'll do a test today.

Thank you so much, @cdent!

antonlindstrom:
@cdent I can verify that this works as expected, thanks!

arioch:
Thanks @antonlindstrom & @cdent!

cdent:
\o/

PR #26, opened 2015-02-12
Support notification script

Spredzy:
This commit adds support for notification script. This allows a script to take action upon sentinel event (ex. +odown, +sdown, +switchmaster).
It also fixes minor lint issues.


arioch:
Looks good.
Thanks!

PR #28, opened 2015-02-20
redis::config inherits redis

systemsathomesdotcom:
This allows the template variables to be scoped locally.  Also
amended the test to look for the default port number in the
config file template output.

With no parameters, on Ubuntu 12, the redis.conf file was coming up with no parameters.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -2,7 +2,7 @@
 #
 # This class provides configuration for Redis.
 #
-class redis::config {
+class redis::config inherits redis {
   if $::redis::notify_service {
     File {
       owner  => $::redis::config_owner,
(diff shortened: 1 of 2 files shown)


arioch:
Inheritance will be deprecated in Puppet 4.0.0.
I'd prefer not to use it anymore.

systemsathomesdotcom:
Ok, well then I can mod the template, switching the ```<%= @varname %>``` to ```<%= scope['::redis::varname'] %>```.  Would that work?

arioch:
I'm guessing that should be ok but it breaks older puppet versions (anything below 3.0.0) which I sadly still need to maintain.
Also I'm concerned why it doesn't seem to be in scope already...

arioch:
Should be fixed in in #40 https://github.com/arioch/puppet-redis/commit/edd7cb55931fe0336bfee475c738ac4b91308f98.

PR #31, opened 2015-05-05
added log_dir_mode param

christofhaerens:


arioch:
Looks like it's b0rking on puppetlabs/apt version 2.0.x.

If you could re-add $config_dir_mode I'll merge it right away.

manifests/config.pp
```diff
@@ -30,7 +30,7 @@
     $::redis::log_dir:
       ensure => directory,
       group  => $::redis::service_group,
-      mode   => $::redis::config_dir_mode,
```
arioch: Just add $log_dir_mode instead of renaming $config_dir_mode.

christofhaerens: Ja ik ben de log_dir_mode in init.pp vergeten.
Kga die toevoegen, pushen en zal nieuwe pull request doen

christofhaerens:
ok this should be it ..

arioch:
Almost there.
```spec/classes/redis_spec.rb``` remains untouched in your follow-up commits. :)

**Comment by christofhaerens** [comment]
updated redis_spec.rb

File: manifests/params.pp [hunk 2]
```diff
@@ -78,6 +78,7 @@
       $config_group              = 'root'
       $config_owner              = 'root'
       $daemonize                 = true
+      $log_dir_mode              = '0755'
```
arioch: It's better to move this outside ```case $::osfamily {}```.
There's no point in repeating the same line with the exact same value for each ```$::osfamily```.

christofhaerens:
It was first outsode $osfamily. (my first commit) But the default in the sles rpm provided by suse is to have 750. So to be able to add Suse support in a later ull request I need to define 750 for Suse. And cannot overwrite it in the params.pp. So I moved it to the osfamily (commit 2).

arioch:
Oh ok, in that case leave it as is.

arioch:
Ok final request before I merge: could you rebase it to a single commit?

christofhaerens:
4 commits rebased into a single

arioch:
Thanks!
Merged upstream.

christofhaerens:
thxs for the help. Up to my next pull request :-)

arioch:
Looking forward to merging it!

PR #33, opened 2015-05-06
Suse conf file fix

christofhaerens:
conf file name must match pid file name otherwise service status and stop doesn't work


christofhaerens:
thxs! Had missed it in the initial tests

arioch:
btw feel free to add SLES to metadata.json

christofhaerens:
even if the sentinel is untested?

arioch:
Good point, that should be tested first.

PR #35, opened 2015-06-01
fix issue with params.pp with strict_variables enabled

eoly:
When strict_variables are enabled Puppet runs fail with Evaluation Error: Unknown variable: '::redis::params::ppa_repo'. 

This happens when the osfamily fact is set to anything but Debian because the ppa_repo variable is not defined and is referenced in init.pp parameters.

Defining the variable under the other osfamilys and setting it to undef fixes the issue.

https://docs.puppetlabs.com/references/latest/configuration.html#strictvariables


arioch:
Thanks for catching this!

PR #40, opened 2015-06-19
access out-of-scope variables via the scope.lookupvar method

eoly:
The templates/redis-sentinel.conf.erb and templates/redis.conf.erb are accessing out-of-scope variables by accessing variables declared top level redis class.

In Puppet 4 and Puppet 3 with Future Parser enabled the values of these variables are blank in the resulting configuration files.

Puppet 4 and Puppet 3 with Future Parser enabled no longer supports relative namespace lookups.
https://tickets.puppetlabs.com/browse/PUP-121

You must access these variables by calling the scope.lookupvar function with the fully qualified variable name as the argument. This PR modifies the 2 templates to use the scope.lookupvar function.


arioch:
You've saved me the hassle of doing it myself. :+1: 
Thanks!

PR #43, opened 2015-06-26
Params set to undef are being included in the redis.conf file as a literal 'undef'

dmcnicks:
When parameters are explicitly set to `undef`, the if statements in the `redis.conf.erb` template file evaluates them as true, which means that those parameters are inserted as a literal `undef`. For example:

    syslog-facility undef

I have updated the if statements that test variables that are set to undef by default to use `defined?`, which properly returns 'false' when the parameter being tested is set to `undef`. For example:

    <% if defined? @syslog_facility %>syslog-facility <%= @syslog_facility %><%     else %># syslog-facility local0<% end %>

I found this issue running puppet agent on a Debian Jessie vagrant box. I have not tested my fork on any other operating systems or versions. I have also updated the relevant if statements in `redis-sentinel.conf.erb` but I have not installed sentinel myself.

### DIFF
--- a/templates/redis-sentinel.conf.erb
+++ b/templates/redis-sentinel.conf.erb
@@ -7,10 +7,10 @@ sentinel monitor <%= @master_name %> <%= @redis_host %> <%= @redis_port %> <%= @
 sentinel down-after-milliseconds <%= @master_name %> <%= @down_after %>
 sentinel parallel-syncs <%= @master_name %> <%= @parallel_sync %>
 sentinel failover-timeout <%= @master_name %> <%= @failover_timeout %>
-<% if @auth_pass -%>
+<% if defined? @auth_pass -%>
 sentinel auth-pass <%= @master_name %> <%= @auth_pass %>
 <% end -%>
-<% if @notification_script -%>
+<% if defined? @notification_script -%>
 sentinel notification-script <%= @master_name %> <%= @notification_script %>
 <% end -%>
 
--- a/templates/redis.conf.erb
+++ b/templates/redis.conf.erb
@@ -60,7 +60,7 @@ logfile <%= @log_file %>
 # syslog-ident redis
 
 # Specify the syslog facility.  Must be USER or between LOCAL0-LOCAL7.
-<% if @syslog_facility %>syslog-facility <%= @syslog_facility %><% else %># syslog-facility local0<% end %>
+<% if defined? @syslog_facility %>syslog-facility <%= @syslog_facility %><% else %># syslog-facility local0<% end %>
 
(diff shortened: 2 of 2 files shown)


eoly:
@dmcnicks can you provide the version of puppet you are running?

According to https://docs.puppetlabs.com/guides/templating.html#testing-for-undefined-variables the instance variable should have a value of nil if it is explicitly set to undef. Therefore,

    <% if @syslog_facility %>

should resolve to nil when explicitly set to undef. This is the case when I test on version 3.8.1 of puppet.

dmcnicks:
> According to https://docs.puppetlabs.com/guides/templating.html#testing-for-undefined-variables the instance variable should have a value of nil if it is explicitly set to undef. Therefore,
> 
> <% if @syslog_facility %>
> 
> should resolve to nil when explicitly set to undef.

> This is the case when I test on version 3.8.1 of puppet.

I am using version 3.7.2 on Debian Jessie installed using the Puppetlabs apt package.

Just checking again. If I revert to release 1.1.0 in my Puppetfile and run the puppet agent I get this in the output:

Error: /Stage[main]/Redis::Service/Service[redis-server]: Failed to call refresh: Could not restart Service[redis-server]: Execution of '/usr/sbin/service redis-server restart' returned 1: Job for redis-server.service failed. See 'systemctl status redis-server.service' and 'journalctl -xn' for details.
Error: /Stage[main]/Redis::Service/Service[redis-server]: Could not restart Service[redis-server]: Execution of '/usr/sbin/service redis-server restart' returned 1: Job for redis-server.service failed. See 'systemctl status redis-server.service' and 'journalctl -xn' for details.
Wrapped exception:
Execution of '/usr/sbin/service redis-server restart' returned 1: Job for redis-server.service failed. See 'systemctl status redis-server.service' and 'journalctl -xn' for details.

And in the redis.conf file:

root@rails:~# grep undef /etc/redis/redis.conf 
syslog-facility undef
slaveof undef
masterauth undef
maxmemory-policy undef
maxmemory-samples undef
include undef
root@rails:~#

If I jump forward to add my changes again and rerun the agent it completes successfully, starts redis and in the redis.conf file:

root@rails:~# grep undef /etc/redis/redis.conf 
root@rails:~#

David

eoly:
If you are running 1.1.0 there was a bug that has since been fixed. 

See ef0fadfe99cc88a508a0fe8054975fa3d9f8745f for a fix.

Can you trying pulling down the latest version of the module from github and test?

arioch:
This has been open quite a while now.
Feel free to reopen this ticket of create a new one if @eoly's (thanks again btw!) patch doesn't solve this exact problem.

PR #46, opened 2015-08-04
Save db to disk

adrian-balcan-ygt:
I added save_db_to_disk parameter.


arioch:
Thanks for your contrib!
I've added a unit test for this parameter: fcf53708b26b66c5d4465112afe66d8e9558d6bd.

PR #48, opened 2015-08-18
Bump puppetlabs-stdlib version spec

### LINKED ISSUE #47: Unable to bring in as a dependency
I've tried adding this to my Modulefile, also tried calling it in my Puppetfile but librarian-puppet always returns "Cannot resolve the dependencies". Is there a problem with this module on the forge?

gblair:
Fixes #47


arioch:
Excellent, thanks for catching this.

PR #50, opened 2015-08-21
add support for hz option

nerzhul:
Redis support hz option to improve performance of the background tasks.


arioch:
Looks good, thanks!

PR #55, opened 2015-09-10
use versioncmp() to check RedHat/CentOS/SCL version

TheMeier:
comparing CentOS version strings wich look like '7.1.1503' gives an error

"Could not retrieve catalog from remote server: Error 400 on SERVER:
comparison of String with 7 failed at modules/redis/manifests/preinstall.pp:10"

### DIFF
--- a/manifests/preinstall.pp
+++ b/manifests/preinstall.pp
@@ -7,7 +7,7 @@ class redis::preinstall {
   if $::redis::manage_repo {
     case $::operatingsystem {
       'RedHat', 'CentOS', 'Scientific', 'OEL': {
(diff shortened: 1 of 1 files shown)


v6:
Does this work with older versions of Facter? 

Also, @TheMeier, does this pull request fix the problem you mentioned in the first comment?

TheMeier:
Nothing about facter/facts was changed here and versioncmd is available for many years in puppet https://projects.puppetlabs.com/issues/1807 and yes this fixes the error

arioch:
Merged.
Thanks!

PR #56, opened 2015-09-21
Sentinel daemonize

Vincent--:
Add "daemonize" option to redis sentinel
Allow to use sentinel class without redis class on a node


Vincent--:
On my setup, I use sentinel class without redis class. With that setup, the daemonize option was set to "no" by default in the sentinel config file and the service never restart...
With this patch, everything is now working properly.

I didnt update travis as the sentinel test includes :
```
  let :pre_condition do
    [
     'class { redis: }'
    ]
  end
```
and then never fails...

I think this pre-condition should be removed as sentinel node don't have to install redis-server.

arioch:
Merged.
Thanks!

PR #63, opened 2015-10-12
Add a option to override the service provider

nerzhul:
When using this module on Debian 8, redis-server try to use the init script instead of the systemd unit file.
This change permit to overwrite service provider


arioch:
Thanks!

PR #65, opened 2015-10-15
Default $daemonize to true on RedHat OS. Fixes #64

### LINKED ISSUE #64: $daemonize is defaulted to 'false' on Redhat OS, which causes service command to hang
The `$daemonize` param is defaulted to [false](https://github.com/arioch/puppet-redis/blob/master/manifests/params.pp#L110) for RedHat based OS'. 
This causes `service redis start` to hang indefinately. 

PR incoming to fix default value on RedHat OS'.

fatmcgav:
Default $daemonize to true on RedHat OS. Fixes #64

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -107,7 +107,7 @@ class redis::params {
       $config_file_mode          = '0644'
       $config_group              = 'root'
       $config_owner              = 'redis'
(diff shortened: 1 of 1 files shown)


arioch:
Merged.
Thanks!

PR #66, opened 2015-10-26
Fix Support for EL7 and Puppet 4

trlinkin:
This PR fixes up the support for EL7 systems in a manner that works on both Puppet 3 and Puppet 4.

It does however come with the cost of no longer supporting Facter below version 1.7.0 or Puppet Enterprise below release 3.0.0, however these are old enough that this should not be a concern.

This PR does duplicate some of the work in #55 but aims to span the major Puppet versions and have coherent testing additions.


arioch:
Looking good, thanks!

PR #70, opened 2015-11-26
Fixing typo on variable substitution on sentinel.pp

### LINKED ISSUE #69: Typo on sentinel.pp on if defined
No variable substitution on:
https://github.com/arioch/puppet-redis/blob/master/manifests/sentinel.pp#L174
Because the variable was enclosed within single quotes.

I will make a PR to solve this.

marcgascon:
This gives an error when installing sentinel if you have already installed the redis-server package:
```
Duplicate declaration.....
```
**resolves #69

### DIFF
--- a/manifests/sentinel.pp
+++ b/manifests/sentinel.pp
@@ -171,7 +171,7 @@ class redis::sentinel (
 ) inherits redis::params {
   $daemonize = $::redis::daemonize
 
-  unless defined(Package['$package_name']) {
+  unless defined(Package["$package_name"]) {
     ensure_resource('package', $package_name, {
       'ensure' => $package_ensure
     })


marcgascon:
Hello I don't know why the tests are not passing with that change, but can you accept my PR @arioch?? Thanks!!

v6:
I think you have to modify the tests, @marcgascon.

v6:
// , Otherwise, it's probably a bad idea to merge this.

manifests/sentinel.pp
```diff
@@ -171,7 +171,7 @@
 ) inherits redis::params {
   $daemonize = $::redis::daemonize
 
-  unless defined(Package['$package_name']) {
+  unless defined(Package["$package_name"]) {
     ensure_resource('package', $package_name, {
```
arioch: Both ways are wrong.
This should do the trick: ```unless defined(Package[$package_name]) {```

arioch:
Merged & fixed.
Thanks!

PR #82, opened 2016-01-13
change redis.conf directory for RedHat systems

tobiasschaber:
Hi,
I would like to see the redis.conf in a sub directory of /etc, not in /etc itself.

I think there is no reason to keep the file up there.

Kind regards,

tobias

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -111,7 +111,7 @@ class redis::params {
     'RedHat': {
       $config_dir                = '/etc/redis'
       $config_dir_mode           = '0755'
-      $config_file               = '/etc/redis.conf'
+      $config_file               = '/etc/redis/redis.conf'
       $config_file_mode          = '0644'
       $config_group              = 'root'
       $config_owner              = 'redis'


v6:
// , Are you thinking of something like /etc/redis.d/ ? 

Also, shouldn't you update the test suite? 

This apparently breaks the build.

petems:
@tobiasschaber the redis.conf file in RedHat is `/etc/redis.conf` in the provided RPM:

```
%config(noreplace) %attr(0644, root, root) "/etc/logrotate.d/redis"
%config(noreplace) %attr(0644, redis, root) "/etc/redis-sentinel.conf"
%config(noreplace) %attr(0644, redis, root) "/etc/redis.conf"
```

If you wish to change it, 

Changing the default param would be an breaking change to the module.

If you wish to provide a different path, you would just call the class with a different path:

```puppet
class { 'redis':
  config_file => '/etc/redis/redis.conf',
}
```

As far as I can see from the module and reading the redis docs, there's no need for a config dir at all _unless_ you're moving it from it's default location at /etc/redis.conf

petems:
Closing as default does not need to be changed, can be given as parameter

PR #83, opened 2016-01-29
creating a seperate copy of redis.conf to allow sentinel to make changes

jacobmw:
When sentinel detects a failover it reconfigures redis.conf to change the 'slaveof' setting.  Puppet was then changing it back to the initial setting.  This uses the same configuration as sentinel.pp had in place.


jacobmw:
This travis build is failing in puppet 4 but with a file I did not modify.  It is failing on preinstall.pp.

arioch:
That's a creative solution to the problem but I don't see any other way myself.
Thanks for the contribution! :+1:

arioch:
Merged.

PR #85, opened 2016-02-18
Marc/redis server and sentinel both running or not

marcgascon:
The idea to solve the issue is something like that. I have to test it but I've created the PR @arioch just in case anyone wants to participate on it.


marcgascon:
Now I've tested on debian and it's working properly. I'm not so proud of the if statements on init.pp so if you have any idea of how to structure it, it will be more than welcome, but the issue is solved with that.

Hope to recive some feedback soon.

Thanks in advance.

arioch:
My apologies for the late response...

Looks good though.
Could you rebase that onto the latest commit?

marcgascon:
Thanks @arioch now branch is rebased. You are able to merge it, but if you can take another look before, that would be great!

Thanks!

petems:
@marcgascon There's been a few merges with this one, can you re-open with a rebase if still needed?

PR #89, opened 2016-03-18
Make notification of service optional

michaeltchapman:
Prevent catalog failures in cases where service_manage is
not enabled


arioch:
Excellent, thanks!

PR #98, opened 2016-06-15
move workdir settings from global section as a database on FreeBSD pl…

olevole:
…atform placed on different place


olevole:
two fix for FreeBSD platform

olevole:
Another impove for FreeBSD: change pidfile as usual for FreeBSD  users

PR #102, opened 2016-07-20
Sentinel Bind

sam-ez4x:
Redis 3.2 introduced '[protected-mode](http://redis.io/topics/security#protected-mode)' to enhance security.   The feature introduces an error in the sentinels when trying to connect from non loopback interface : 

> DENIED Redis is running in protected mode because protected mode is enabled, no bind address was specified, no authentication password is requested to clients....

Fix adds 'bind' to the sentinel template in order to allow the sentinels to communicate.

```
class { 'redis::sentinel':
  sentinel_bind => '0.0.0.0',
}
```

Example tested in production, Centos 6.5, Redis 3.2.1.  Configuration, Master / Slave plus multiple Client Sentinels.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -43,6 +43,7 @@ class redis::params {
   $requirepass                     = undef
   $save_db_to_disk                 = true
   $sentinel_auth_pass              = undef
+  $sentinel_bind                   = undef
   $sentinel_config_file_mode       = '0644'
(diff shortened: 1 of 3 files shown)


mnaser:
Welp, just noticed this PR exists, just submitted a similar one #109 with tests.. Be nice if this (or the other) get merged.

PR #109, opened 2016-09-11
Allow binding port for Sentinel

mnaser:
The protected mode of Redis will prevent the Sentinel from starting up
if it is not bound to a specific interface.  This allows configuring
sentinel to be bound to an interface.


petems:
Closed via #120

PR #127, opened 2016-10-14
Move Repo for RHEL systems to EPEL

### LINKED ISSUE #103: powerstack.org - No longer hosted
powerstack.org appears to be taken down. Are there other providers for the RPM's for REHL based distros?

petems:
* powerstack.org no longer exists
* Redis is now in EPEL

Closes #103

### DIFF
--- a/manifests/preinstall.pp
+++ b/manifests/preinstall.pp
@@ -6,52 +6,8 @@
 class redis::preinstall {
   if $::redis::manage_repo {
     case $::operatingsystem {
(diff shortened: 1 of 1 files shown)


v6:
// , I think EPEL only has ReDiS 2 for CEntOS 6, right? 

Anyway, IUS appears to host a copy of ReDiS 3.2 for CEntOS 6: 

https://dl.iuscommunity.org/pub/ius/stable/Redhat/6/x86_64/repoview/letter_r.group.html

PR #131, opened 2016-11-04
Add systemd service file template

gabrielecerami:
If the unix socket is configured to be present under a dir in a tmpfs,
like /var/run/redis the directory must be created before launching
redis. The best way to do it is inside service file, ensure the
directory is present (and eventually security context restored) each
time redis is started.

### DIFF
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -4,6 +4,12 @@
 #
 class redis::service {
   if $::redis::service_manage {
+    file {
+      $::redis::service_file_orig:
+        ensure  => present,
+        content => template($::redis::service_template);
+        mode    => $::redis:service_file_mode
+    }
     service { $::redis::service_name:
       ensure     => $::redis::service_ensure,
       enable     => $::redis::service_enable,
--- a/templates/redis.service.erb
+++ b/templates/redis.service.erb
@@ -1,0 +1,16 @@
+[Unit]
+Description=Redis persistent key-value database
+After=network.target
+
+[Service]
+ExecStartPre=/usr/bin/mkdir -p --mode=<%= @unixsocketperm %> <%= File.dirname(@unixsocket) %>
+ExecStartPre=/usr/bin/chown redis:redis <%= File.dirname(@unixsocket) %>
+ExecStartPre=-/usr/sbin/restorecon -Rv <%= File.dirname(@unixsocket) %>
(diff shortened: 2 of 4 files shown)


templates/redis.service.erb
```diff
@@ -0,0 +1,16 @@
...
+Description=Redis persistent key-value database
+After=network.target
+
+[Service]
+ExecStartPre=/usr/bin/mkdir -p --mode=<%= @unixsocketperm %> <%= File.dirname(@unixsocket) %>
+ExecStartPre=/usr/bin/chown redis:redis <%= File.dirname(@unixsocket) %>
+ExecStartPre=-/usr/sbin/restorecon -Rv <%= File.dirname(@unixsocket) %>
+ExecStart=/usr/bin/redis-server /etc/redis.conf --daemonize no
+ExecStop=/usr/libexec/redis-shutdown
+User=redis
```
petems: This should probably be the `$service_user` variable rather than hard coded

templates/redis.service.erb
```diff
@@ -0,0 +1,16 @@
...
+Group=redis
```
petems: This should probably be the `$service_group` variable rather than hard coded

manifests/service.pp
```diff
@@ -4,6 +4,12 @@
 #
 class redis::service {
   if $::redis::service_manage {
+    file {
+      $::redis::service_file_orig:
```
petems: Wouldn't this mean that the systemd unit file is always added, even if the user is not on a systemd system?

templates/redis.service.erb
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Redis persistent key-value database
+After=network.target
+
+[Service]
+ExecStartPre=/usr/bin/mkdir -p --mode=<%= @unixsocketperm %> <%= File.dirname(@unixsocket) %>
+ExecStartPre=/usr/bin/chown redis:redis <%= File.dirname(@unixsocket) %>
```
petems: `$service_user`:`$service_group`

gabrielecerami:
I realized this may not be the best way to deal with tmpfs. I'll close this pull request and create another that uses systemd-tmpfiles instead

petems:
@gabrielecerami Cool, no problem! 👍

PR #132, opened 2016-11-29
Refactor

codylane:
This pull request addresses the _pending_ tests.  Still probably needs some more work but I didn't want to spend all day on this.  Please let me know if you have any questions or concerns.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -115,6 +115,13 @@ class redis::config {
         owner  => $::redis::config_owner,
       }
 
+      file { '/var/run/redis':
+        ensure => 'directory',
+        owner  => 'redis',
+        group  => 'redis',
+        mode   => '0755',
+      }
+
       if $::redis::ulimit {
         augeas { 'redis ulimit' :
           context => '/files/etc/default/redis-server',
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ doc
 log/
 pkg
 spec/fixtures
+.bundle/
--- a/.rspec
+++ b/.rspec
@@ -1,0 +1,4 @@
+--format d
+--color
+--tty
+--backtrace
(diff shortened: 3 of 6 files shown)


codylane:
Yay! All tests passed now.

manifests/config.pp
```diff
@@ -115,6 +115,13 @@
         owner  => $::redis::config_owner,
       }
 
+      file { '/var/run/redis':
+        ensure => 'directory',
+        owner  => 'redis',
```
petems: This needs to be `$::redis::config_owner`

manifests/config.pp
```diff
@@ -115,6 +115,13 @@
...
+        group  => 'redis',
```
petems: This needs to be `$::redis::config_group`

petems:
@codylane Also, could you open an issue for the Debian failure? I can't see that issue when running the Beaker tests

manifests/config.pp
codylane: Ah yes, my bad, I meant to fix that yesterday.  Let me do that right quick.

codylane:
@petems - Thank for your review and for helping remind me that I need to still fix a few things. 

The /var/run/redis issue only seemed to affect the Ubuntu 16.04 nodeset.  It took me a while to try and duplicate it on my own because I have a feeling something changed in the `redis-server` package for Ubuntu based hosts.  I'm happy to open a separate issue, but it appears that you already did?

petems:
@codylane I ended up breaking the PR into seperate chunks whilst I tested something: #135 and #134 

Yeah, it looks like the Ubuntu package is working, as the Ubuntu Beaker test works ok. The Redis one might actually be an upstream bug with the run directory, I'm investigating... but regardless, all working now, tests are green, so lets close this one! 👍

codylane:
Awesome!  I'm happy to hear that this helped refactor some of the tests and what not.  This module is awesome!  I have a few other things I'd like to add, but i'll do that in a separate request.  Thank you so much for your quick review and for the feedback.  Let's keep making this thing better.

PR #136, opened 2016-11-30
Adds redis-server version fact

petems:
* Will be useful in the future to only allow 3.3+ config on newer packages
* Also useful for seeing the version of the redis package easily


codylane:
Awesome.  I was just thinking about adding this. ;)

PR #139, opened 2016-11-30
for folks that do not use redis to cache to disk

codylane:
This pull requests addresses a possible use case (at least for my team) that doesn't use redis to cache to disk.  I'm hoping that this can be merged to master where it's a noop for anyone that doesn't use redis the way my team does, but gives my team the ability to not cache to disk.  

Please let me know if i've missed anything.  I appreciate your time and feedback.

### DIFF
--- a/spec/classes/redis_spec.rb
+++ b/spec/classes/redis_spec.rb
@@ -231,6 +231,16 @@ describe 'redis', :type => :class do
         }
       end
 
+      describe 'without parameter dbfilename' do
+        let(:params) {
+          {
+            :dbfilename => false,
+          }
+        }
+
+         it { is_expected.to contain_file(config_file_orig).without_content(/^dbfilename/) }
+       end
+
       describe 'with parameter hash_max_ziplist_entries' do
         let (:params) {
           {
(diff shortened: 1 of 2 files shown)


petems:
LGTM, when Travis goes green I'll merge! 👍

codylane:
You rock dude, thank you so much.

codylane:
Well, that failure is interesting, I cannot duplicate this on my workstation.  I've spent an hour now trying to understand what is going on with the single failing test and I'm seriously scratching my head.

@petems - I know you are working on this too.  Do you have any ideas?

codylane:
Thinking about this deeper, this feels more like a ruby bug.  Let me see if I can find a bug report about this.

petems:
I believe it's actually an issue with how undef is represented in older Puppet

codylane:
Ahh, that also makes sense.  Can you help me understand which version of puppet is being installed in travis for that failing test?  I apologize if this is easy to find and I'm overlooking it.

petems:
I think it should be fixed in this, it's weird issues with how `scope[` works with `undef` values

codylane:
Ugh, so we have to go through all the templates and add the following to test for undef?  Would you like me to do that?  I have no problem doing so.

```
unless [:undef, false, nil, ''].include?(scope['somevar'])
```

codylane:
I wish we could just use this but I can see how it could be problematic for folks as I've seen some people quote booleans, or quote undef... etc.

```
<% unless scope['somevar'].nil? -%>
# do something when variable is defined
<% else -%>
# do something else when variable is not defined
<% end -%>
```

PR #147, opened 2016-12-06
Added Archlinux as supported OS

bartjanssens92:
This commit adds the needed params for Archlinux.


bartjanssens92:
I've changed the parameter warnings.

petems:
I've tested this on an Arch VM and it works as itended! 👍 

Merging 😄

bartjanssens92:
All right, thnx :)

PR #149, opened 2016-12-07
Archlinux: Added tests and update config_dir parameter

bartjanssens92:
This commit adds tests for Archlinux.
While testing found an issue with the 'config_dir' parameter.

Signed-off-by: bjanssens <bjanssens@inuits.eu>


bartjanssens92:
Thnx!

petems:
@bartjanssens92 👍

Issue #152, opened 2017-02-19
Deprecate Puppet 3.X Support


petems:
Puppet 3.X is EOL

So, we want to upgrade this module to have Puppet 4 support.

- [ ]  Create a last release tag in the major version
- [x]  Add notes to the README/add warning to the code for the deprecation
- [ ]  Make a new release, and add types to parameters (Puppet 4 only feature)
- [ ]  Update module to use the new Puppet Lookup/Data in modules work from Puppet 4 onwards

v6:
// , How will those of us poor folk still stuck with Puppet 3 continue to use this? Shall we just install the older version that supports Puppet 3?

v6:
// , If we do use that older version, can we be assured that it will use versions of its dependencies that also work with Puppet 3?

petems:
There will be a final tag for the Puppet 3 features, so you can fork from that tag and continue to support Puppet 3 from that tag. 

The dependancy modules are pinned to versions that support Puppet 3, so the other module dependancies shouldn't be an issue:

```json
{"name":"puppetlabs/apt","version_requirement":">= 2.3.0 <3.0.0"},
{"name":"puppetlabs/stdlib","version_requirement":">= 1.0.2 <5.0.0"},
{"name":"stahnma/epel","version_requirement":">= 1.0.2 <2.0.0"}
```

I would recomend upgrading to Puppet 4 however 😄

v6:
// , Establishing good habits of "software life cycle management", e.g. not using End of Life infrastructure, with people who don't want it can be a struggle, sometimes.

petems:
Done in https://github.com/voxpupuli/puppet-redis/commit/f14f07db1be7b1ded2cd7ac233d43bdd5e3681db

v6:
Is that the "final tag" you alluded to earlier?

PR #153, opened 2017-02-19
Adding note about Puppet 3 support

petems:
* Adding deprecation warning
* First steps towards the work in #152

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,18 @@ Test against a specific Puppet or Facter version:
     $ PUPPET_VERSION=4.10.0 bundle update && bundle exec rake spec
     $ FACTER_VERSION=1.6.8  bundle update && bundle exec rake spec
 
+## Puppet 3 Support
+
+Puppet 3 is EOL as-of January 2017. The last release of this module that will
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -94,6 +94,17 @@ Test against a specific Puppet or Facter version:
     $ PUPPET_VERSION=2.7.19 bundle update && bundle exec rake spec
     $ FACTER_VERSION=1.6.8  bundle update && bundle exec rake spec
 
+## Puppet 3 support
+
+Puppet 3 is EOL as-of January 2017. The last release of this module that will
+support Puppet 3.X and earlier will be 1.2.5
```
ghoneycutt: Suggest removing the exact version and just say that the v1 series will support Puppet3 and v2 will support Puppet4.

petems: Yeah, meant to edit this, was planning on bumping this version to 3.0.0 (so it matches the Puppet version 😄)

Releasing is cheap, and we've done a number of changes to the module we did since the 1.2.X version could possibly be breaking changes, so I thought might as well bump it up 👍

PR #162, opened 2017-04-25
Changes for RHEL-ish specific configuration

petems:
* Previously we were changing some of the default 
configurations from the upstream packages
* This causes issues with CentOS 6 services, and 
needlessly clutters logs.
* Change to follow ownership patterns of 
original packages:
  * Changes service group to be root
  * Change workdir permission
  * Change pid file location back to package default
* Adds failing Travis check for CentOS 6


v6:
// , YES. Thank you. This actually goes along better with certain groups hardening guidelines, too.

PR #163, opened 2017-04-25
  // , Add currently parameterized values into 3.2's redis.conf

v6:
This simply takes the redis.conf file from Redis 3.2, and
adds our parameterized variables into it.

I used https://raw.githubusercontent.com/antirez/redis/3.2/redis.conf
as a starting point, and copied and replaced as needed.

If this pull request looks good, I'll start adding some parameters, defaults, and Ruby templating code in here, `params.pp`, and `init.pp` for the values we don't yet have parameters for in this module.

What do you think?

--Nathan Basanese


v6:
// , Does this conflict with https://github.com/arioch/puppet-redis/pull/161/commits/b55d2494f53b4e132309c1ece359615927d775cc from https://github.com/arioch/puppet-redis/pull/161 ?

v6:
// , @petems please let me know if I should rebase on top of the code for https://github.com/arioch/puppet-redis/pull/161 and make a pull request to `petems:fix_centos_package_changes`.

Maybe that way I'm not getting in your way. :)

petems:
@v6 Yeah, I started on some of those changes in that file, it's still a work in progress.

The main issue is that a lot of the newer config settings will break older redis, such as the `protected-mode` config setting 😢

v6:
// , Hm. This is only going to get more "interesting" with ReDiS 4.0.

We could give the parameters such as the `protected-mode` a default value of `undef`, and require that you only set it if you are using a later version. Seems hacky, though.

Maybe the puppet-redis module could handle configuration from later versions that breaks older Redis with a parameter or fact for the Redis version number?

Is it time to do something like `redis.conf-2.x.erb`, `redis.conf-3.0.erb`, `redis.conf-3.2.erb`, and `redis.conf-4.0.erb`?

I'm not too familiar with how to make a Puppet module configure multiple backwards incompatible versions of an application in Puppet, honestly.

I have to think someone else has run into it, though.

What do you think about this? Is there a common "pattern" out there for dealing with this type of problem? 

Also, on a side note, I wonder what, if anything, this module might do if someone's running multiple Redis on a single host, and any Redis has a different version than any other on that host.

petems:
Yeah, what makes it even harder is that there are breaking changes even on patch level changes:

2.6.0
2.8.0
2.8.13
2.8.15
2.8.5
2.8.9
3.0.0
3.2.0

all have config settings that are backwards incompatible... which would lead to soooo many config templates required for each version. 😢

v6:
// , I posted about this on everyone's favorite question and answer site, if you'd like to weigh in: 

https://stackoverflow.com/questions/43646190/how-do-i-handle-changing-backwards-incompatibility-of-software-versions-an

v6:
// , @petems, all existing checks pass.

If I go with option 2 and option 3 from the following post, would you mind pointing me to a list of backwards incompatible configuration options that 3.2 brings?

If so, I may add checks for them to `spec/classes/redis_spec.rb`.

v6:
// , By the way, about that idea of having so many config templates, isn't that sort of what went down here? 

https://github.com/arioch/puppet-redis/pull/145/files#diff-310d9735fce6b39d468e45471fda1c88

v6:
// , @petems, how about a compromise?

1. The module could have a separate `.erb` Ruby template file for each minor version, not each patch version. Since these are not released all that frequently, it would mean a new `.erb` Ruby template file every few months, at most, regardless of how many changes occur in the mean time to Redis.

2. For backwards incompatible changes that occur with the patch level changes, the module could handle them with logic within the module or templates. 
Perhaps this could happen by leaving the defaults for the backwards incompatible parameters as **`undef`** in `params.pp`. Given that it's a patch level change, I don't think it would end up  too  hairy. 
Then, the README would have to note that these must be specified manually if a user happens to have a later patch version of Redis.  **OR** we could add a parameter for the ReDiS version to `manifests/params.pp`.

It would mean 7 templates for the year:
`redis.conf.2.6.erb`
`redis.conf.2.8.erb` (if else blocks handle backwards incompatible changes among `2.8.13`, 2.8.15, 2.8.5, and 2.8.9)
`redis.conf.3.0.erb`
`redis.conf.3.2.erb`
`redis.conf.4.0.erb`
`redis.conf.4.2.erb`
`redis.conf.4.4.erb`

@petems , what do you think?

v6:
// , My plan to make this compatible with https://github.com/arioch/puppet-redis/commit/49425a5515fccbe56c820f990464e8f635a73874#diff-0ac866172efd29d0ddc2aa75ae6f182dR88 : 

1. Add parameters for the configuration values in https://raw.githubusercontent.com/antirez/redis/3.2/redis.conf to params.pp
2. Create `templates/redis.conf.3.2.8.erb` from https://raw.githubusercontent.com/antirez/redis/3.2/redis.conf , this time parameterizing ALL configuration options for ReDiS 3.2.8.
3. Ensure backwards compatibility, through logic in the `templates/redis.conf.3.2.8.erb` template, with prior versions of ReDiS 3.2.
4. Add an option for `redis/redis.conf.3.2.8.erb` to `manifests/config.pp` at https://github.com/arioch/puppet-redis/blob/master/manifests/config.pp#L95, similar to the option above it.
5. Make a new pull request
6. Fix broken tests
7. Add more automated tests for the new parameters (I may need advice on this one).

v6:
// , The new redis.conf.3.2.erb will need to make sure the following are parameterized: 

parameterize these: 

protected-mode yes

supervised no

syslog-ident undef

rdbchecksum yes

repl-diskless-sync no

appendfsync everysec

lua-time-limit 5000

cluster-slave-validity-factor undef

cluster-migration-barrier undef

cluster-require-full-coverage undef

list-max-ziplist-size -2

list-compress-depth 0

client-output-buffer-limit normal 0 0 0
client-output-buffer-limit slave 256mb 64mb 60
client-output-buffer-limit pubsub 32mb 8mb 60

v6:
// , I looked at every commit associated with `redis.conf` in https://raw.githubusercontent.com/antirez/redis/3.2/00-RELEASENOTES, and it looks like none of the changes are backwards incompatible. As far as I can tell from looking at that briefly, it doesn't look like I need to add any logic to the template.

v6:
// , Maybe I don't step 4, with a new pull request. I'm going to try steps 6 & 7, now, it seems.

templates/redis.conf.3.2.erb
```diff
@@ -0,0 +1,1087 @@
...
+#
+# The server only accepts connections from clients connecting from the
+# IPv4 and IPv6 loopback addresses 127.0.0.1 and ::1, and from Unix domain
+# sockets.
+#
+# By default protected mode is enabled. You should disable it only if
+# you are sure you want clients from other hosts to connect to Redis
+# even if no authentication is configured, nor a specific set of interfaces
+# are explicitly listed using the "bind" directive.
+protected-mode <% if @protected-mode -%>yes<% else -%>no<% end -%>
```
petems: This needs to be protected_mode, variables in Puppet are alphanumeric + underscores

v6: // , Ouch, how could I miss that?!?

v6: // , Rebased in the correction, will double check the other vars tonight.

desertkun:
@v6 @petems  Would like to report an issue: when installing redis, it sets the `protected-mode` conf variable even if the version of the redis is 2.8:

image

Second launch fixes the problem:

```
Notice: /Stage[main]/Redis::Config/Redis::Instance[default]/File[/etc/redis/redis.conf.puppet]/content: content changed '{md5}15d9d3f2553bb77e4df6f4989f40b5f5' to '{md5}d86d6d999a495c8e6a1dfed9b9bfd094'
Notice: /Stage[main]/Redis::Config/Redis::Instance[default]/Exec[cp -p /etc/redis/redis.conf.puppet /e
```
Apparently, the module somehow detects the 3.2 version of redis, because there is no one, and then catch up.

petems:
@desertkun What os is this happening on?

desertkun:
Debian 8. I can send you a whole virtual machine state (virtual box) to reproduce the case.

ekohl:
This now needs a serious rebase. Due to its age, I'm going to close this now but a fresh PR is more than welcome.

PR #164, opened 2017-04-25
Adds specific versions to fixtures

petems:


.fixtures.yml
```diff
@@ -1,9 +1,14 @@
 fixtures:
   repositories:
-    apt: 'git://github.com/puppetlabs/puppetlabs-apt.git'
-    stdlib: 'git://github.com/puppetlabs/puppetlabs-stdlib.git'
-    epel: 'https://github.com/stahnma/puppet-module-epel.git'
-
+    apt:
```
v6: // , This seems like a good idea. 

You never know when a change there might break something here!

PR #167, opened 2017-04-28
Update config file

v6:
// , This works backward from the configuration file for Redis 3.2. 

What do you think?

More docs and discussion here: https://github.com/arioch/puppet-redis/pull/163

--Nathan Basanese


v6:
// , I made this to the wrong fork.

PR #176, opened 2017-05-01
add systemd support for sentinel service

KlavsKlavsen:
I based the redis-sentinel service file on Ubuntu's - and made it settable using the params that was already in the module.

It is a bit odd that redis module includes init script for redis-sentinel -but not for redis.
The norm is to NOT include init scripts.. I assume that on some distro's redis-sentinel had no init script, so that would be why it has one for redis-sentinel but not for redis (especiall now that redis-sentinel is part of same redis binary from 3.0+ :)

Looking forward to comments/suggestions - I tried to keep it the same way as the rest of the code.. restraining myself from any puppet-4-isms :)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -641,6 +641,6 @@ class redis (
   exec { 'systemd-reload-redis':
     command     => 'systemctl daemon-reload',
     refreshonly => true,
-    path        => '/bin:/usr/bin:/usr/local/bin'
+    path        => '/bin:/usr/bin:/usr/local/bin',
   }
 }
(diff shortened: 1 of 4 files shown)


petems:
Looks like there's a dependency loop? 

```
(Exec[systemd-reload-redis] => Class[Redis] => Class[Redis::Sentinel] => File[/etc/systemd/system/redis-sentinel.service] => Exec[systemd-reload-redis])
```

KlavsKlavsen:
odd.. not in the puppet 4 setup I tested on. I'll have a look

petems:
Rebased in https://github.com/arioch/puppet-redis/pull/180

PR #179, opened 2017-05-03
Add redisget()

ghoneycutt:
Add the redisget() function which can get a value for a specified key
from a specified redis URL.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -32,6 +32,8 @@ group :test do
   gem 'puppet-syntax',                                              :require => false, git: 'https://github.com/gds-operations/puppet-syntax.git'
   gem 'pry'
   gem 'rb-readline'
+  gem 'redis',                                                      :require => false
+  gem 'mock_redis',                                                 :require => false
 end
 
 group :system_tests do
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,5 @@
    Copyright (C) 2012 Tom De Vylder
+   Copyright (C) 2017 Garrett Honeycutt <code@garretthoneycutt.com>
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
--- a/README.md
+++ b/README.md
@@ -67,6 +67,17 @@ With adjustments:
       failover_timeout => 30000,
     }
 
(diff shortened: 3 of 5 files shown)


petems:
@ghoneycutt This is very cool! Let me have a look, I might add and acceptance test also, but otherwise this seems like an awesome addition 👍

ghoneycutt:
Thanks, glad you like it. These tests were a pain :>

Could you please tag a new release?

petems:
Will do when the acceptance tests are merged 👍

petems:
@ghoneycutt BTW do you have some example usecases of how you're using this right now in your Puppet code you can share? 

It'd be a good bit of background for the README, also thinking about adding a `redisset()` command in Puppet also, but that might not be so great for idempotency

arioch:
@petems How does this work with the Forge? Do you have publish access on my account?

petems:
@arioch opened an issue #185

ghoneycutt:
I have a CD pipeline for a piece of software that records the version of a successful build in our source of truth (Redis). To know what version should be installed, I use redisget() similar to the example in the README and use that in my package resource.

Another use case is similar to Consul for service discovery. You can set that information in Redis using a process outside of Puppet and then have Puppet query that to know what information to use in your configuration.

A more detailed piece of code might look like

```puppet
class profile::myapp {
  $redis_url = hiera('redis_url')

  $version = redisget("version.myapp.${::environment}", $redis_url)

  package { 'myapp':
    ensure => $version,
  }
}
```

ghoneycutt:
I don't have a use case for Puppet to set data in Redis as I'm using Redis as source of truth to be queried. I don't really see a use case to write data there with Puppet. Suggest holding off on writing additional functions until someone actually needs them, even if they are trivial to implement.

petems:
Yep, all makes sense 👍

I'll hold fire until someone asks for something like that, but seems like a cool usecase 😄

v6:
// , This doesn't break compatibility with CEntOS 6's Ruby 1.8.7, correct?

ghoneycutt:
@v6 I'm using it in a Puppet v3 / Ruby 1.8.7 environment

PR #180, opened 2017-05-03
Add systemd support for sentinel service

petems:
Rebased of #176 with some spec fixes

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -637,4 +637,10 @@ class redis (
       fail "Replication is not possible when binding to ${::redis::bind}."
     }
   }
+
+  exec { 'systemd-reload-redis':
+    command     => 'systemctl daemon-reload',
+    refreshonly => true,
+    path        => '/bin:/usr/bin:/usr/local/bin'
+  }
 }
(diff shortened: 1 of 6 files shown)


petems:
@KlavsKlavsen Ok, so this seems to work for Ubuntu, but the service file for redis breaks on CentOS (as the acceptance tests show)

The default service file in CentOS looks like this:

```
[Unit]
Description=Redis Sentinel
After=network.target

[Service]
ExecStart=/usr/bin/redis-sentinel /etc/redis-sentinel.conf --daemonize no
ExecStop=/usr/bin/redis-shutdown redis-sentinel
User=redis
Group=redis

[Install]
WantedBy=multi-user.target
```

This works with no issues, so there's something in the new config that causes issues.

I put logic around the `/bin/run-parts` because they're Debian-flavour only, but still having issues.

KlavsKlavsen:
its already odd that you want to control the service/init scripts for redis-sentinel - and NOT for redis.. :)
I assume its a legacy from when there was no "redis-sentinel" service - and you wanted to support that ?
we'd need a centos version in the patch and select based on os.

petems:
Yeah, the packages for Ubuntu 14.04 and Debian stable don't have the redis-sentinel package. This module dates back to 2014, so there's some stuff in here that's probably not as relevant now. 

I'm just double checking from your issue before, are you using the ppa for redis or the standard xenial repo?

KlavsKlavsen:
standard xenial repo..

petems:
We might want to manage this in the future but for now I think we're ok leaving this, as the original location of the config file was fixed in #184  so closing

PR #181, opened 2017-05-03
Adds acceptance tests for the redisget() function

petems:
Extended From #179


ghoneycutt:
Thanks for adding these tests!

PR #183, opened 2017-05-03
(testing) Simplify command run by TravisCI

ghoneycutt:
Instead of anding all the commands together, we can pass them each to
bundle exec, which achieves the same results.


petems:
Seems like a good idea to me, if it's green I'll merge 👍

Issue #105, opened 2016-08-30
Add overcommit.pp to deal with `Can't save in background: fork: Cannot allocate memory` ? 


v6:
I ran into this, recently: 

```
    [beta1redis213] out: 17662:M 28 Aug 12:39:14.595 # Can't save in background: fork: Cannot allocate memory
```

Turns out it's a common enough problem that someone put it in the ReDiS FAQ, and in somebody stuck a solution for it in their Puppet module for ReDiS: 

ReDiS FAQ on overcommit: http://redis.io/topics/faq#background-saving-is-failing-with-a-fork-error-under-linux-even-if-i39ve-a-lot-of-free-ram

https://github.com/giosakti/puppet-redis/blob/master/manifests/overcommit.pp

Is this over-engineering, or is it worth looking in to?

petems:
Closed via #188

PR #188, opened 2017-05-03
Adds redis::administration class

petems:
* Allows modification of common redis kernel
tweaks and requirements
* The ones that antirez (redis creator) mentions
and that are mentioned specifically in redis 
when running it in the logs are:
  * Disable THP: https://redis.io/topics/latency
  * Enable Overcommit Memory: https://redis.io/topics/admin
  * somaxconnL http://download.redis.io/redis-stable/redis.conf


petems:
Closes #105

Issue #185, opened 2017-05-03
Deployment to Forge


petems:
@arioch So we have a few choices for this:

One, we add something to the .travis.yml like this:

```
deploy:
  provider: puppetforge
  user: puppet
  password:
    secure: "KOhDZN2IcdlCbnz4i6wCbh1RXK/TZW7GTzxCwuM4wnSVTit/I9TpD2AdEIHCYg3D7V0BKYqvKhVl0zVlabinHltDf0l9FJ0ic9yfXF3M2gcvJtOKjn+Xm8rC+UDFeiaIByDjE6bc7/Ys0SBQq/yLXkjI+OShqhtdo26ZIXiPVGo="
  on:
    tags: true
    # all_branches is required to use tags
    all_branches: true
    # Only publish the build marked with "DEPLOY_TO_FORGE"
    condition: "$DEPLOY_TO_FORGE = yes"
```

This will auto deploy everytime a build is succesful and the tag is changed.

In the mean time, I might just ask you do do a 2.0.0 tag release soon because setting up the above will be a PITA in the short term (but much better in the long term! 😄 )

PR #191, opened 2017-05-03
Make TravisCI push to the Forge

arioch:
As suggested in #185.
Why not bite the bullet and just try. 😄


petems:
Ok, lets get this working! 👍

PR #196, opened 2017-05-04
Specify sort key in redis.conf.erb

sshipway:
This adds Ruby 1.8 compatibility which was lost in v1.2.4


petems:
@sshipway Thanks! Will merge when green 👍

petems:
Rebased with additions to the 2.4.10 and 2.8.0 erb files in #199

PR #198, opened 2017-05-05
Adds tests for when Redis service is unmanaged

### LINKED ISSUE #197: The fix for issue #192 broke service_managed false
I have the following hiera values:
redis::managed_by_cluster_manager": true
redis::service_manage": false
redis::notify_service": false

But now an 'include ::redis' gives me:
Error: Could not find dependent Service[redis] for Augeas[Systemd redis ulimit] at /etc/puppet/modules/redis/manifests/ulimit.pp:44

Did you mean to make https://github.com/arioch/puppet-redis/blob/master/manifests/ulimit.pp#L31 an 'elsif' and not an 'if'?

petems:
* If the redis service is unmanaged, like so:
```
redis::managed_by_cluster_manager": true
redis::service_manage": false
redis::notify_service": false
```
then you get a compilation failure as it tries to
look for the redis service to notify:
```
But now an 'include ::redis' gives me:
Error: Could not find dependent Service[redis] for Augeas[Systemd redis ulimit] at /etc/puppet/modules/redis/manifests/ulimit.pp:44
```
* In fact, we don't need any service notification 
in the redis::ulimit class, as it's already in the
redis::config class, which notifies the service 
class anyways
* Adds spec for this use-case to avoid regression

Closes #197

### DIFF
--- a/manifests/ulimit.pp
+++ b/manifests/ulimit.pp
@@ -50,7 +50,6 @@ class redis::ulimit {
         "set \$nofile/value \"${::redis::ulimit}\""],
       notify  => [
         Exec['systemd-reload-redis'],
-        Service[$::redis::service_name]
       ],
     }
   } else {
--- a/spec/classes/redis_ulimit_spec.rb
+++ b/spec/classes/redis_ulimit_spec.rb
@@ -16,6 +16,36 @@ describe 'redis::ulimit' do
         }'
       ]
     end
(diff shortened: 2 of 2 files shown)


mbaldessari:
Tested this and everything worked correctly. Thanks again for the quick fix!

petems:
@mbaldessari Excellent! 🚢 🇮🇹

Issue #195, opened 2017-05-04
Sort problem in v1.2.4 template


sshipway:
In version 1.2.4, the template has:
```
<% if @save_db_to_disk %>
<%- @save_db_to_disk_interval.sort_by{}.each do |seconds, key_change| -%>
save <%= seconds -%> <%= key_change -%> <%= "\n" -%>
<%- end -%>
<% end %>
```
This throws up an error under our version of Ruby (1.8.7):
```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Evaluation Error: Error while evaluating a Function Call, Failed to parse template redis/redis.conf.erb:
  Filepath: /etc/puppet/environments/development/modules/redis/templates/redis.conf.erb
  Line: 117
  Detail: undefined method `<=>' for nil:NilClass
 at /etc/puppet/environments/development/modules/redis/manifests/config.pp:92:18 on node xmdrdb01.smxemail.com
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run
```
The solution is to specify what to sort by, then it will correctly run:
```
<% if @save_db_to_disk %>
<%- @save_db_to_disk_interval.sort_by{|k,v|k}.each do |seconds, key_change| -%>
save <%= seconds -%> <%= key_change -%> <%= "\n" -%>
<%- end -%>
<% end %>
```
In the meantime, we have reverted to v1.2.3 which runs as expected and provides the features we currently use.

petems:
@sshipway Can you open a PR to the repo with that change? 

I'm assuming this is on CentOS 6? 1.8.7 is EOL. I'd recomend using the Puppet 4+ agent as it comes with it's own Ruby packaged in, so won't have Ruby version issues.

sshipway:
Yes, this is Centos 6 with Puppet 3.  We have a project running to upgrade to Centos 7 and bring everything up to Puppet 4, but this will take much time before it is approved and finalised.  In the meantime, we (and many like us) will continue to be using Centos6/Pupet3 etc I suspect.

I don't have a separate repo with a change, sorry; it's only 6 chars and we're using the official 1.2.3 tag.  I'll see what I can do...

petems:
Closed via #199

PR #199, opened 2017-05-05
Update sort to specify key

petems:
* With the previous approach, error on 1.8.7 Ruby:
```
Error: Failed to parse template redis/redis.conf.2.4.10.erb:
  Filepath: /etc/puppet/modules/redis/templates/redis.conf.2.4.10.erb
  Line: 87
  Detail: undefined method `<=>' for nil:NilClass
 at /etc/puppet/modules/redis/manifests/config.pp:91 on node master
```
* For ruby 1.8 compatibility
* Should fix older Puppet on systems that have 
1.8.7 Ruby (RHEL6/CentOS6)
* We're not officially supporting 1.8.7 as it's 
EOL, but since this is a minor change, isn't too 
much of a maintenance cost, lets do it! :dancer:

### DIFF
--- a/templates/redis.conf.2.4.10.erb
+++ b/templates/redis.conf.2.4.10.erb
@@ -84,7 +84,7 @@ databases <%= @databases %>
 #
 #   Note: you can disable saving at all commenting all the "save" lines.
 <% if @save_db_to_disk %>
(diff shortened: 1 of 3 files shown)


petems:
Closes #196 #195

Issue #113, opened 2016-09-22
Support for multi-instances per host


rhoml:
It is a common practice to have more than one instance of redis per host. We should move the config.pp into a define type so we can create multiple instances pointing into different ports. 

Thoughts?

spo0nman:
Yes! I support this. I am actually in the process of hacking this into the module. maybe we can collaborate.

rhoml:
Absolutely, maybe rolling out small changes into a dev branch?

spo0nman:
@rhoml do you have a branch already that i can chekout?

arioch:
I'd prefer having an $instance parameter in the main class which the config class could read, parse and deploy using a resource defined type.

Pseudo code:

```
class redis (
  $instance = {},
) {
}

class redis::config {
  create_resources(::redis::instance, $::redis::instance, {})
}

define redis::instance(
  $instance_name,
  $instance_config_dir,
  $instance_config_file,
  $instance_ip,
  $instance_port,
) {
}
```

rhoml:
@spo0nman  No I don't have any working branch atm, been busy. 

@arioch yeah, I'd like to see something similar to elastic/puppet-elasticsearch https://github.com/elastic/puppet-elasticsearch/blob/master/manifests/instance.pp

spo0nman:
@arioch how would we make sure that all the instances are individually configurable? e.g if redis::config is the top layer, and multiple instances on a single host will need instance specific overrides for max_memory etc.

arioch:
@spo0nman by using ```redis::config``` for its original purpose: to configure generic stuff in ```/etc/redis```.

Instances could have their own directory structure, f.i. ```/opt/redis/instance_1``` managed by ```redis::instance```.

spo0nman:
@arioch the updated one looks good. let me write something around this idea.

spo0nman:
Hi, 

small update: I have a decent working version now with these changes. I had to make a few changes to accomplish this. I have a rather large patch, I'll commit it to my fork for your review tomorrow.

petems:
@spo0nman Hey, @arioch has added me to the repo to help with maintenance, I look forward to your PR! 👍

spo0nman:
Hello!

I have pushed[1] my changes to my fork. It is a large change, I took the idea of @arioch and created a define called redis::instance, which is called multiple times from redis::config.  

I added a new variable called instances, this is expected to be a hash of the form: 
`$instances = { 'instance_name' = { 
[...]

PR #200, opened 2017-05-06
Add redis::instance defined type

petems:
This will allow configuring multiple redis instances on one machine, allowing the first step towards #113

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -3,70 +3,6 @@
 # This class provides configuration for Redis.
 #
 class redis::config {
-  $activerehashing               = $::redis::activerehashing
-  $aof_load_truncated            = $::redis::aof_load_truncated
-  $aof_rewrite_incremental_fsync = $::redis::aof_rewrite_incremental_fsync
-  $appendfilename                = $::redis::appendfilename
-  $appendfsync                   = $::redis::appendfsync
-  $appendonly                    = $::redis::appendonly
-  $auto_aof_rewrite_min_size     = $::redis::auto_aof_rewrite_min_size
-  $auto_aof_rewrite_percentage   = $::redis::auto_aof_rewrite_percentage
-  $bind                          = $::redis::bind
-  $cluster_config_file           = $::redis::cluster_config_file
-  $cluster_enabled               = $::redis::cluster_enabled
-  $cluster_node_timeout          = $::redis::cluster_node_timeout
-  $daemonize                     = $::redis::daemonize
-  $databases                     = $::redis::databases
-  $dbfilename                    = $::redis::dbfilename
-  $extra_config_file             = $::redis::extra_config_file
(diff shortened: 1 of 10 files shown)


petems:
@jacobmw @v6 @spo0nman @bostrowski13

Working MVP of two redis instances on one server, running on two different ports:

```puppet
class { '::redis':
      manage_repo     => #{manage_repo},
      default_install => false,
    }

    redis::instance_config {'redis1':
      port     => '7777',
      pid_file => '/var/run/redis/redis-server-redis1.pid'
    } ~> redis::instance_service {'redis1':}

    redis::instance_config {'redis2':
      port     => '8888',
      pid_file => '/var/run/redis/redis-server-redis2.pid'
    } ~> redis::instance_service {'redis2':}
```

Does this fit people's needs? I'm planning on merging the instance_config and instance_service into one to make this a bit easier to use 😄

bostrowski13:
this work sfor me.  are other params defined within each instance like you specified the pid_file?

petems:
@bostrowski13 Yep, all of the parameters defined are the same that would be the defined the main class. In fact, you no longer need to specify the pid, you can just do:

```
class { '::redis':
      default_install => false,
    }

    redis::instance {'redis1':
      port     => '7777',
    }

    redis::instance {'redis2':
      port     => '8888',
    }
```

bostrowski13:
that is excellent.  thank you!

spo0nman:
@petems good to know that this code is getting used. In my earlier PR there was a bug where the data directory for redis was shared between the instances, I fixed it in my fork but never published. Look out for this bug. Good luck!

petems:
@spo0nman Ooh, interesting, that's a good point actually. I should probably add a test to make sure each instance uses it's own data folder

But yes, your original code was instrumental in getting this working so thank you very much! 😄

tdotgreenshirt:
What about having separate sentinel instances? Is it possible to do this or can sentinel monitor multiple Redis instances without creating a separate sentinel instance for each.

petems:
@tdotgreenshirt From my understanding, you don't want the sentinel instance on the same machine as the redis instance because it would be a SPOF: When you lose that host you lose both.

If there is a case for multi-sentinal on one host, then it could be added as a feature?

tdotgreenshirt:
@petems This is similar to what we are doing:
https://i.stack.imgur.com/Citcm.png

However, if we have separate Redis instances, we also need a separate Sentinel instances.

petems:
@tdotgreenshirt Ok, can you add that as an issue and I'll tag it as an upcoming feature.

Now we have a defined instance type, and redis-sentinel is just another redis instance, this should be fairly quick to implement 😄

bostrowski13:
thats actually very similar as to what we're after as well.  be great to have that option for sentinel.

tdotgreenshirt:
Created issue for multiple Sentinel instances: https://github.com/arioch/puppet-redis/issues/223

PR #202, opened 2017-05-17
Ignore selinux default context for /etc/systemd/system/redis.service.d

amoralej:
When creating new files/directory/symlink, puppet asumes that it's a file
to get the default context. This is wrong as the default context depends
on the file type. See https://tickets.puppetlabs.com/browse/PUP-7559

For the redis case, when creating /etc/systemd/system/redis.service.d
it's setting the context redis_unit_file_t, which is wrong. It should
be systemd_unit_file_t. This has two efects, it's setting the directory
to a wrong context on first run and it's breaking the idempotency of
puppet-redis.

With this patch, puppet-redis is ignoring selinux context for the directory.
Note that the default context is applied automatically by the OS, so it
will be set to the right one.

### DIFF
--- a/manifests/ulimit.pp
+++ b/manifests/ulimit.pp
@@ -30,9 +30,10 @@ class redis::ulimit {
   }
   if $service_provider_lookup == 'systemd' {
     file { "/etc/systemd/system/${::redis::service_name}.service.d/":
-      ensure => 'directory',
-      owner  => 'root',
-      group  => 'root',
+      ensure                  => 'directory',
+      owner                   => 'root',
+      group                   => 'root',
+      selinux_ignore_defaults => true,
     }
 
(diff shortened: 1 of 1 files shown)


petems:
@amoralej I will admit I'm not super well-versed on Selinux, but wouldn't it be better to change this to specify the seltype parameter on the file to `systemd_unit_file_t` instead?

amoralej:
@petems it'd work specifying seltype also, but i prefer not to use a parameter to force to default value which will be applied by the OS. I think it's unnecessary and may be a problem if the required value is changed in future or if required context is different in other OS versions (i'd say it's unlikely in this case, but still possible).

petems:
Ok, seems reasonable.

Thanks for your contribution! 👍

Issue #150, opened 2017-01-11
Ownership problem with Ubuntu redis-server


jerith:
On Ubuntu 14.04, `redis-server` has an init script that does `chown redis:redis $RUNDIR $PIDFILE` (where `RUNDIR=/var/run/redis` and `PIDFILE=$RUNDIR/redis-server.pid`) at startup. This causes puppet to "fix" the ownership of `/var/run/redis` (to `redis:root` unless `config_owner` and `config_group` have been set differently), which triggers a restart of `redis-server`, which changes the ownership of `/var/run/redis`, etc. forever.

Is it possible to tell puppet not to care about the owner or group of `/var/run/redis` beyond the initial creation?

petems:
I think we can fix that easier by just changing the `$config_group` default on Ubuntu to Redis.

Can you try changing that in the code in your Ubuntu 14.04 machine and see if it changes anything else? So add a section in the params to change  `config_group` in Ubuntu 👍

jerith:
My workaround was to set `redis::config_group` to `redis` in hiera, which seems to have sorted out the problem.

jerith:
On debian jessie (which we also run), a stock install of `redis-server` (from jessie-backports, although I don't think that makes a difference) has both `/var/run/redis` and `/etc/redis/redis.conf` owned by `redis:redis`. `/etc/redis` and everything else inside it is owned by `root:root` with world-readable permissions.

nvtkaszpir:
The same happens on debian wheezy.

dominics:
> Can you try changing that in the code in your Ubuntu 14.04 machine and see if it changes anything else?

Can confirm (albeit with 16.04) that it changes the group of `/etc/redis`, `/etc/redis/redis.conf` and `/etc/default/redis-server`, not just `/var/run/redis`. Fine for me (I want to `CONFIG REWRITE` anyway) - but, in general, not isolated to `/var/run`.

PR #203, opened 2017-05-18
Add switch to manage File[/var/run/redis] on debian systems

ntnn:
The existence of /var/run/redis causes the startup of redis-server via
systemd to fail on debian systems with redis-server >= 3.2.8-2.

The default unit file for redis-server in debian jessie-backports
contains `RuntimeDirectory=redis`, see [this commit](https://github.com/lamby/pkg-redis/commit/1cecea5abcb2ce05beacd4347977634d06c59cef).

This causes systemd to attempt to create `/var/run/redis`, which fails,
since puppet already created it:

```
May 18 14:12:12 po-redis-gitlab-qa-bs01 systemd[1]: Starting Advanced key-value store...
May 18 14:12:12 po-redis-gitlab-qa-bs01 systemd[24311]: Failed at step RUNTIME_DIRECTORY spawning /bin/run-parts: File exists
May 18 14:12:12 po-redis-gitlab-qa-bs01 systemd[24312]: Failed at step RUNTIME_DIRECTORY spawning /usr/bin/redis-server: File exists
May 18 14:12:12 po-redis-gitlab-qa-bs01 systemd[1]: redis-server.service: control process exited, code=exited status=233
May 18 14:12:12 po-redis-gitlab-qa-bs01 systemd[1]: Failed to start Advanced key-value store.
May 18 14:12:12 po-redis-gitlab-qa-bs01 systemd[1]: Unit redis-server.service entered failed state.
May 18 14:12:12 po-redis-gitlab-qa-bs01 systemd[1]: redis-server.service holdoff time over, scheduling restart.
May 18 14:12:12 po-redis-gitlab-qa-bs01 systemd[1]: Stopping Advanced key-value store...
```

Additionally the `$var_run_redis_mode` is wrong as well, since redis-server
sets it to `0755` and the module changes it back to `2755`.

Also see issue #150, the same applies here - puppet changes the mode of
`/var/run/dir` and refreshes `Service[redis-server]`, which then fails
to restart because systemd can't create the directory.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -135,20 +135,22 @@ class redis::config {
         owner  => $::redis::config_owner,
       }
 
-      case $::operatingsystem {
-        'Debian': {
-          $var_run_redis_mode = '2775'
+      if $::redis::manage_var_run_dir {
+        case $::operatingsystem {
+          'Debian': {
+            $var_run_redis_mode = '2775'
+          }
+          default: {
+            $var_run_redis_mode = '0755'
+          }
         }
-        default: {
-          $var_run_redis_mode = '0755'
-        }
-      }
 
-      file { '/var/run/redis':
-        ensure => 'directory',
-        owner  => $::redis::config_owner,
-        group  => $::redis::config_group,
-        mode   => $var_run_redis_mode,
+        file { '/var/run/redis':
+          ensure => 'directory',
+          owner  => $::redis::config_owner,
+          group  => $::redis::config_group,
+          mode   => $var_run_redis_mode,
+        }
       }
 
     }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -520,6 +520,11 @@
 #
 #   Default: 5000
 #
+# [*manage_var_run_dir*]
+#   Manage /var/run/dir on debian systems.
+#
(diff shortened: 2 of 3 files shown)


petems:
Rebased in #204 to fix tests

ntnn:
That change, as it has been merged, will actually break the module for other users using debian jessie without backports. The unit file there is still a simple one without Pre/PostRun and RuntimeDirecotry directives.

I went with the switch instead of detection because it is not feasible to attempt to properly detect a backports- and default version.

petems:
The acceptance tests run with the package from Jessie and are working fine, can you show me the error?

ntnn:
Huh, I guessed it would break compatibility because of the difference in the unit file, which for default jessie hosts (without backports) is this: 
```
[Unit]
Description=Advanced key-value store
After=network.target

[Service]
Type=forking
ExecStart=/usr/bin/redis-server /etc/redis/redis.conf
ExecStop=/usr/bin/redis-cli shutdown
Restart=always
User=redis
Group=redis

[Install]
WantedBy=multi-user.target
```

But if the tests worked fine I'll withdraw that - thanks for the quick merge!

PR #207, opened 2017-05-18
Add an optional third parameter to redisget() to specify a default value

ghoneycutt:


ghoneycutt:
@petems I got the acceptance tests this time :)

petems:
@ghoneycutt Awesome stuff! 😄 

I might ask you to rebase as I was doing some work to update the function docs to work with puppet-strings (see #208), but overall looks good

lib/puppet/parser/functions/redisget.rb
```diff
@@ -3,20 +3,40 @@
...
+      default = args[2]
+      raise(Puppet::ParseError, "redisget(): Wrong argument type given (#{default.class} for String) for arg3 (default)") if default.is_a?(String) == false
+    end
 
-    redis = Redis.new(:url => url)
-    redis.get(key)
+    raise(Puppet::ParseError, "redisget(): Wrong argument type given (#{key.class} for String) for arg1 (key)") if key.is_a?(String) == false
+    raise(Puppet::ParseError, "redisget(): Wrong argument type given (#{url.class} for String) for arg2 (url)") if url.is_a?(String) == false
+
+    begin
```
petems: This logic looks a little confusing to me, could we clean this up somehow?

lib/puppet/parser/functions/redisget.rb
```diff
@@ -3,20 +3,40 @@
...
+      redis = Redis.new(:url => url)
+      k = redis.get(key)
+      if k == nil and defined?(default) != nil
+        default
+      else
+        k
+      end
+    rescue
```
petems: Is there a way to pick which error to rescue from here?

ghoneycutt:
No problem, I'll rebase once #208 gets merged.

lib/puppet/parser/functions/redisget.rb
petems: Something like 

```
rescue Exception => e
      if default
        debug "Connection to redis failed with #{e}"
        default
      else
        raise(Puppet::Error, "Connection to redis server failed - #{e}")
      end
    end
```

Gives a little more detail for both the happy and sad paths

lib/puppet/parser/functions/redisget.rb
petems: Possibly something like changing `k` to `returned_value`? And is there a way to break out the nesting? Not sure if there's something better it just looks a little confusing IMO

ghoneycutt: changed the variable name and added the exception message to the output as requested

petems:
Merged in #209

PR #214, opened 2017-07-11
updated redis systemd unit file for better use with instances

bostrowski13:
made some small updates to the systemd service unit file to better handle multiple instances.


petems:
@bostrowski13 Awesome, LGTM! 😄

PR #215, opened 2017-07-11
Use package_ensure if it specifies a version instead of the minimum_version

sp-joseluis-ledesma:
When using package_ensure => some_version, it will initially configure the the redis using the minimum_version variable anyway instead of taking in account the package_ensure variable. This PR fixes that.

### DIFF
--- a/manifests/instance.pp
+++ b/manifests/instance.pp
@@ -161,6 +161,7 @@ define redis::instance(
   $no_appendfsync_on_rewrite     = $::redis::no_appendfsync_on_rewrite,
   $notify_keyspace_events        = $::redis::notify_keyspace_events,
   $managed_by_cluster_manager    = $::redis::managed_by_cluster_manager,
+  $package_ensure                = $::redis::package_ensure,
   $port                          = $::redis::port,
   $rdbcompression                = $::redis::rdbcompression,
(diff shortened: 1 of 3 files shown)


petems:
Ah, good point, didn't think of that usecase! Let me review this 😄

manifests/instance.pp
```diff
@@ -276,7 +277,11 @@
     refreshonly => true,
   }
 
-  $redis_version_real = pick(getvar_emptystring('redis_server_version'), $minimum_version)
+  if $package_ensure =~ /^[0-9]\.[0-9]/ {
```
petems: Hmm, we might want to extend this regex. If someone's using an rc/alpha candidate this regex won't match (eg. https://github.com/antirez/redis/archive/4.0-rc3.tar.gz)

Can we extend the regex and add some extra usecases to the specs? eg. 3 digits (3.2.1), non-numeric characters (4.0.0-rc)

sp-joseluis-ledesma: The regex matches (I did not look for the end of the regex, only start with [0-9]\.[0-9]). Anyway I've improved it a bit (/^[0-9]+\.[0-9]/) so it will work when redis get to version 10 :).

I've added a couple more of specs usecases.

petems:
Awesome, LGTM! 😄

PR #217, opened 2017-07-12
Split the redis instance logfiles by default

kwevers:
Avoid the logs of multiple instances going into the default log file without the user having to define a separate log file for each instance


petems:
@kwevers Looks good to me, can you add a spec for the instance defined type showing it uses a unique logfile by default?

kwevers:
@petems I've added some test. 
Let me know if this is not what you expected

petems:
@kwevers Perfect! Merging 😄

PR #219, opened 2017-07-12
Split Redis workdir

kwevers:
Make sure that each Redis instance uses a separate workdir.
This avoids instances overwriting files of other instances.


petems:
@kwevers Thanks! 😄

PR #220, opened 2017-07-13
Split Redis instance socket files

kwevers:
Define separate socket files for the Redis instances by default.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -34,6 +34,7 @@ class redis::config {
       pid_file            => $::redis::pid_file,
       log_file            => $::redis::log_file,
       manage_service_file => $::redis::manage_service_file,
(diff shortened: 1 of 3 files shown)


kwevers:
I think this is the last change required to have a sane default config for multiple instances.
The only thing the users needs to specify is the port

kwevers:
@petems The Ruby 2.0 tests fail because of a dependency issue:

```
Gem::InstallError: faraday-http-cache requires Ruby version >= 2.1.0.
An error occurred while installing faraday-http-cache (2.0.0), and
Bundler cannot continue.
```
I'm not very knowledgeable in this area. Could you perhaps take a look at this?

petems:
@kwevers Just updated the Ruby versions 👍 Running again

PR #222, opened 2017-07-13
Instance service improvements

kwevers:
Some small fixes to align manifest documentation and file path inheritance

The last commit reworks the naming convention for non default instances:
- makes sure the config file name ends on .conf(.puppet) instead of the instance name.
- include redis in the service name

The service name could be useful for people running multiple apps on the same server with for example dedicated JBoss and Redis instances.
In this way they could have a service jboss-server-app1 and redis-server-app1 instead of a non ambiguous service app1

We could drop the -server part to shorten the names if required.

### DIFF
--- a/manifests/instance.pp
+++ b/manifests/instance.pp
@@ -63,6 +63,7 @@
 # @param [String] save_db_to_disk_interval    save the dataset every N seconds if there are at least M changes in the dataset
 # @param [String] service_enable   Enable/disable daemon at boot.
 # @param [String] service_ensure   Specify if the server should be running.
+# @param [String] service_group   Specify which group to run as.
 # @param [String] service_hasrestart   Does the init script support restart?
 # @param [String] service_hasstatus   Does the init script support status?
 # @param [String] service_user   Specify which user to run as.
@@ -110,6 +111,7 @@
 # @param [String] workdir   The DB will be written inside this directory, with the filename specified
 #   above using the 'dbfilename' configuration directive.
 #   Default: /var/lib/redis/
+# @param [String] workdir_mode   Adjust mode for data directory.
 # @param [String] zset_max_ziplist_entries   Set max entries for sorted sets.
 # @param [String] zset_max_ziplist_value   Set max values for sorted sets.
 # @param [String] cluster_enabled   Enables redis 3.0 cluster functionality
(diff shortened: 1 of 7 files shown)


petems:
This makes sense. Can you fix up the specs? They seem to be Ubuntu only errors, so it's probably something to do with the Debian init file

manifests/instance.pp
```diff
@@ -210,9 +216,18 @@
     $redis_file_name_orig = $config_file_orig
     $redis_file_name      = $config_file
   } else {
-    $redis_config_extension    = ".${title}"
-    $redis_file_name_orig      = "${config_file_orig}${redis_config_extension}"
-    $redis_file_name           = "${config_file}${redis_config_extension}"
+    $redis_server_name    = "redis-server-${name}"
```
petems: If we wanted to be really pedandtic we could change it to be `redis` on RedHat and `redis-server` on Debian to follow convention, but I don't care that much

manifests/instance.pp
```diff
@@ -210,9 +216,23 @@
...
+    $redis_file_name_orig = sprintf('%s/%s.%s', dirname($config_file_orig), $redis_server_name, 'conf.puppet')
+    $redis_file_name      = sprintf('%s/%s.%s', dirname($config_file), $redis_server_name, 'conf')
+  }
+
+  $_real_log_file = $log_file ? {
```
petems: I don't think this is needed, we could simply set the log_file in the instance default declaration, then set the default to be `${log_dir}/redis-server-${name}.log`

kwevers: This is how I initially did it, but some Travis test where failing because the ${log_dir} didn't get replaced.

kwevers:
@petems It took more attempts than I care to admit, but all tests are passing.
Should I look into the redis vs redis-server issue or is it good enough as is?

manifests/instance.pp
petems: Oh yeah, you can't refer to other parameters in parameters.

Ok, leave it like this for now, there's probably a cleaner solution somehow

petems:
@kwevers Your work is appreciated! Can you squash the commits down to one?

We can leave the redis vs redis-server thing, if someone using redis with multiple instances wants to change it they can open a PR, not super important, still works the same functionally 👍

kwevers:
@petems I've rebased it to a few different commits so the different changes are still separated.

petems:
LGTM!

PR #230, opened 2017-07-27
additional updates to support multiple instances

bostrowski13:
Changes include:
addition of redis-shutdown erb template to include path to config file for specific instance.
using the redis-shutdown erb template when systemd is the service controller
change to redis-service erb template to support service name after shutdown command


bostrowski13:
I'm pretty new to this, but I feel like this test for multi-instance isn't quite right.  Looking at the error from Travis...

`Error: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Resource Statement, Duplicate declaration: File[/usr/bin/redis-shutdown] is already declared in file /etc/puppetlabs/code/modules/redis/manifests/instance.pp:276; cannot redeclare at /etc/puppetlabs/code/modules/redis/manifests/instance.pp:276 at /etc/puppetlabs/code/modules/redis/manifests/instance.pp:276:7  at /tmp/apply_manifest.pp.0mWha4:14 on node centos-7-x64.c.travis-ci-prod-2.internal`

Looks like somehow this is getting applied twice, eventhough its only declared once in instance.pp.

Can somebody maybe explain to me whats happening?

petems:
@bostrowski13 If you declare a file that multiple instances are going to use, it's going to lead to duplication. You'll need to move that declaration into the config.pp class.

bostrowski13:
Makes sense now.  I'll fix it up.

bostrowski13:
Alright, seems like the only error i see now is 

```
  Error: Execution of '/opt/puppetlabs/puppet/bin/gem install --no-rdoc --no-ri redis' returned 1: ERROR:  Error installing redis:
  	redis requires Ruby version >= 2.2.2.
```

Maybe this is just an error in the run?

petems:
@bostrowski13 Nah, it looks like the gem requires a newer ruby version, let me fix the test there.

petems:
@bostrowski13 I've got a PR to fix that issue, when it's green I'll merge and it should make this green too 😄 

https://github.com/arioch/puppet-redis/pull/235

bostrowski13:
@petems thanks very much!

petems:
@bostrowski13 Can you rebase with master? Should be all green 👍

bostrowski13:
sorry it took so long, been out at some conferences in the past week or two.

bostrowski13:
so i'm trying to catch up on some of this work.  the last commit added an example to the readme file on how to use instances.  i'd like ot get this merged so i can repull and continue some more work.

templates/service_templates/redis.service.erb
```diff
@@ -1,15 +1,20 @@
...
 After=network.target
+After=network-online.target
+Wants=network-online.target
 
 [Service]
 Type=forking
-ExecStart=/usr/bin/redis-server <%= @redis_file_name %>
-ExecStop=/usr/bin/redis-server -p <%= @port %> shutdown
+ExecStart=/usr/bin/redis-server <%= @redis_file_name %> --supervised systemd
+ExecStop=/usr/libexec/redis-shutdown <%= @redis_server_name %>
```
svs1979: what about something like `ExecStop=/bin/kill -s TERM $MAINPID` instead? This way redis-shutdown isn't needed and afaics Redis handles SIGTERM fine.

svs1979:
As stated inline in the diff, why not use `ExecStop=/bin/kill -s TERM $MAINPID` in the systemd unit to get rid of redis-shutdown?

crispygoth:
FWIW I'm using the module with this PR applied, but without the redis-shutdown script.

In fact, I've removed the ``ExecStop`` entirely, since systemd's default behaviour on stop is to send a SIGTERM anyway, so the explicit kill command isn't really needed: https://github.com/crispygoth/puppet-redis/commit/620599abb460c68c61bac9bd57313a235d72d10c

ekohl:
This now needs a rebase. I'd also prefer a squashed history to just the relevant changes.

ghoneycutt:
Thanks for the contribution! We just added support for multiple instances. If this PR adds even more functionality, please rebase and reopen.

PR #233, opened 2017-08-08
Added configuration options for client-output-buffer-limit

Mike-Petersen:
The configuration for client-output-buffer-limit slave and pubsub can now be configured.


petems:
LGTM! 👍

PR #237, opened 2017-09-11
Bumped apt version dependency to version < 5.0.0

c4m4:


petems:
LGTM! 👍

c4m4:
when the module will be update in puppetforge ?

petems:
@c4m4 Right now we're blocked on https://github.com/arioch/puppet-redis/issues/216, I'm going to ping @arioch and see if we can get that fixed 👍

PR #243, opened 2017-10-09
protected-mode configuration option (Redis 3.2+)

Dan70402:
This allows one to bypass the protected-mode on Redis 3.2+ when listening on all interfaces.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -59,6 +59,7 @@
 # @param [String] package_name   Upstream package name.
 # @param [String] pid_file   Where to store the pid.
 # @param [String] port   Configure which port to listen on.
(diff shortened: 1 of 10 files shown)


petems:
@Dan70402 Awesome, if thise goes green I'll merge 👍

Dan70402:
This was a little more involved than I had first thought.  I found #163 which had some discussion on the best way to manage various breaking changes in the redis configs.  Based on the comments and the existing pattern I implemented the `protected-mode` mode for 3.2.  This will only add the `protected-mode` for 3.2 versions of redis because of the existing string based switch for selecting the redis erb to use.  It should be straight forward to add additional 3.2 configuration if anyone would like.

Issue #227, opened 2017-07-20
Undefined variable 'redis_server_version' warning on first run


benh57:
When this module first runs to install redis, the fact is undefined:  (because it's not installed yet) - Puppet 4 or later warns about this:
 Warning: Undefined variable 'redis_server_version';   (file & line not available)

After the first run, the warning goes away.  To avoid this the fact should be defined to something, even undef might be fine.

TheMeier:
Actuall this leads to errors for me

`Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Function Call, integer 7089836968 too big to convert to `int'`


Any chances to cut a release, theres 24 commits merged already on master an the last release was ~6 month ago :/

petems:
Hi @TheMeier! I'm having issues with releasing right now (See https://github.com/arioch/puppet-redis/issues/262) 

But I'm curious, there's no reason that the value should be that big an integer!

Can you run the following on the machine running that: `redis-server -v` and comment the full output into this issue?

ekohl:
I don't think this is an issue anymore so I'm closing this. Feel free to reopen if it persists.

PR #249, opened 2017-11-12
get rid of getvar_emptystring function

vicinus:
The _getvar_emptystring_ function is a clone of the _getvar_ function from stdlib
with the only functionally difference, that it returns an empty string instead of _undef_.
But _getvar_emptystring_ is always used with pick which handles empty strings and
_undef_ the same way. So there is no need at the moment for _getvar_emptystring_
and even if there is a need in the future it would be better to use something
like:

`(getvar('dummy') == undef) ? { true: '', default: getvar('dummy') }`

The reason is that since the copying of the _getvar_ function to
_getvar_emptystring_ the _getvar_ function was updated to not send warnings
regarding undefined variables, but this update was never adopted into the
_getvar_emptystring_ function. And that is the cause for issue #227


petems:
getvar_emptystring was not an older clone of the getvar from stdlib, it was intentionally made to allow the use of a facter fact that may or may not be set with strict variables set to true. 

This issue only affected Puppet 3 machines, which is now EOL. And a quick test shows that your change works on Puppet 4 machines with strict_variables set to true. So I'm going to merge this 👍

PR #250, opened 2017-12-05
Update redis mode on wheezy

mattkirby:
This commit updates the mode for redis run directory on Debian platforms without systemd. Without this change puppet will attempt to change the mode on /var/run/redis during every puppet run, which triggers a refresh.


petems:
Looks good, will merge when green 👍

petems:
Rebased with specs in #252

PR #254, opened 2017-12-12
Adds logic for installing redid-sentinel package

### LINKED ISSUE #253: Inconsistent sentinel_package_name handling on Debian
In https://github.com/arioch/puppet-redis/blob/6c28bc3/manifests/params.pp#L123  `$sentinel_package_name` is set for all Debian systems to `redis-sentinel`. But in https://github.com/arioch/puppet-redis/blob/2231d4d/manifests/sentinel.pp#L205 the `package` definition happens only for Ubuntu 16.04 and newer (and there's no support for real Debian systems like stretch and sid). This is causing errors on other Debian systems:
```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Invalid relationship: File[/etc/redis/sentinel.conf.puppet] { require => Package[redis-sentinel] }, because Package[redis-sentinel] doesn't seem to be in the catalog
```

I think the condition in sentinel.pp should be moved to params.pp, and the condition in sentinel.pp should be changed to a `unless defined Package[$package_name]` or so.

petems:
* Debian Stretch also added a separate package 
also: https://packages.debian.org/stretch/redis-sentinel
* Closes #253

### DIFF
--- a/manifests/sentinel.pp
+++ b/manifests/sentinel.pp
@@ -200,9 +200,13 @@ class redis::sentinel (
 
   if $::osfamily == 'Debian' {
     # Debian flavour machines have a dedicated redis-sentinel package
-    # This is default in Xenial onwards, unstable debian or PPA/other upstream
+    # This is default in Xenial or Stretch onwards or PPA/other upstream
     # See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=775414 for context
-    if (versioncmp($::operatingsystemmajrelease, '16.04') >= 0 or $::redis::manage_repo) {
+    if (
+      (versioncmp($::operatingsystemmajrelease, '16.04') >= 0 and $::operatingsystem == 'Ubuntu') or
+      (versioncmp($::operatingsystemmajrelease, '9') >= 0 and $::operatingsystem == 'Debian') or
+      $::redis::manage_repo
+      ) {
       package { $package_name:
         ensure => $package_ensure,
       }
--- a/spec/classes/redis_sentinel_spec.rb
+++ b/spec/classes/redis_sentinel_spec.rb
@@ -87,4 +87,26 @@ describe 'redis::sentinel', :type => :class do
     }
   end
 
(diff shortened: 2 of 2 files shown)


manifests/sentinel.pp
```diff
@@ -200,9 +200,13 @@
 
   if $::osfamily == 'Debian' {
     # Debian flavour machines have a dedicated redis-sentinel package
-    # This is default in Xenial onwards, unstable debian or PPA/other upstream
+    # This is default in Xenial or Stretch onwards or PPA/other upstream
     # See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=775414 for context
-    if (versioncmp($::operatingsystemmajrelease, '16.04') >= 0 or $::redis::manage_repo) {
+    if (
+      versioncmp($::operatingsystemmajrelease, '16.04') >= 0 or
```
eserte: This should also include something like `$::operatingsystem == "ubuntu"` and below the same for `"debian"`.

eserte: ... and still I think the condition should go to params.pp. This way it is still broken for debian-jessie and older, as the `require => Package[$package_name];` some lines later will fail otherwise.

petems: Good point, the edgecase is unlikely (Ubuntu on 9 or earlier) but might as well protect against it

petems: The require package name works fine, because on Jessie the package is redis instead of redis-sentinel. The tests work, both unit and acceptance 👍

spec/classes/redis_sentinel_spec.rb
```diff
@@ -87,4 +87,26 @@
     }
   end
 
+  describe 'on Debian Stretch' do
```
eserte: This should probably be "on Debian Jessie".

petems: Fixed! 👍

PR #259, opened 2018-02-17
add automake requirements for gem install

amhammett:
build errors appear to be caused from gem install missing ffi dependencies.


ekohl:
Current master works without so I'm going to close this.

PR #269, opened 2018-08-10
Fix setting ulimit under systemd

bahner:
1.
There was an inconsistency with creating a file
/etc/systemd/system/redis-server.service.d/limit.conf
but then referring to file
/etc/systemd/system/redis-server.service.d/limits.conf
in the augeas resource.

Whereas limits.conf seems to be a better name the file
limit.conf is now likely to exist on many systems, and
I don't want to risk having both a limit.conf and a
limits.conf. So limit.conf it is.

2.
The context was set to to the filepath, but should have been
prefixed with "/files". Seeing as this is default I removed the context.

This closes  arioch/puppet-redis#268

### DIFF
--- a/manifests/ulimit.pp
+++ b/manifests/ulimit.pp
@@ -43,12 +43,12 @@ class redis::ulimit {
       mode   => '0444',
     }
     augeas { 'Systemd redis ulimit' :
-      incl    => "/etc/systemd/system/${::redis::service_name}.service.d/limits.conf",
+      incl    => "/etc/systemd/system/${::redis::service_name}.service.d/limit.conf",
       lens    => 'Systemd.lns',
-      context => "/etc/systemd/system/${::redis::service_name}.service.d/limits.conf",
       changes => [
         "defnode nofile Service/LimitNOFILE \"\"",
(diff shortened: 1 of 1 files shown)


bahner:
Hepp!
The checks that fail in Travis is not something I believe is caused by this pull request. This fails, even on eg. ruby 2.3.3
 	ERROR:  Error installing bolt:
  		winrm-fs requires Ruby version >= 2.2.0.
This seems to be an error with the setup or the winrm-fs plugin.

I will ignore these checks, until I am told otherwise.

Kind regards,
- bahner

bostrowski13:
has this been here for 3 months because of broken ruby dependencies in the travis config?

nicolasbigler:
I would appreciate if this fix can make it to the next release.
Currently the module is essentially broken on servers with sytemd because the limit.conf file is empty due to the typo which causes the systemd unit to be masked and therefore puppet can't enable the service.

arioch:
Merged.
Thx for your patience, there's a vacancy for a new maintainer as Peter stepped down. 😉

PR #270, opened 2018-08-17
Update puppetlabs/apt dependency

l-lotz:
The puppetlabs/apt version 5 adds fixes for Ubuntu 18.04.


l-lotz:
I added a fix for the CI, as the build error was caused by a newer winrm-fs version. I can remove the fix or change it if this is not the correct approach.

arioch:
Only time to actually do so. 😉 

Merged, thanks!

PR #273, opened 2018-09-30
Fixed #246: 2.8 version of Redis on Debian 8 errors out

### LINKED ISSUE #246: 2.8 version of Redis on Debian 8 errors out
From @desertkun:

@petems  Would like to report an issue: when installing redis, it sets the `protected-mode` conf variable even if the version of the redis is 2.8:

image

Second launch fixes the problem:

```
Notice: /Stage[main]/Redis::Config/Redis::Instance[default]/File[/etc/redis/redis.conf.puppet]/content: content changed '{md5}15d9d3f2553bb77e4df6f4989f40b5f5' to '{md5}d86d6d999a495c8e6a1dfed9b9bfd094'
Notice: /Stage[main]/Redis::Config/Redis::Instance[default]/Exec[cp -p /etc/redis/redis.conf.puppet /e
```
Apparently, the module somehow detects the 3.2 version of redis, because there is no one, and then catches up.

desertkun:
On Debian 8, the `apt-policy` still keeps outdated `2.8.17` preferable, so the outdated version gets installed. This fix makes sure the right version will be installed.

See #246

### DIFF
--- a/manifests/preinstall.pp
+++ b/manifests/preinstall.pp
@@ -21,12 +21,11 @@ class redis::preinstall {
             source => 'http://www.dotdeb.org/dotdeb.gpg',
           },
           include  => { 'src' => true },
-          before   => [
-            Class['apt::update'],
-            Package[$::redis::package_name],
-          ],
-        }
-
+        } -> apt::pin { $::redis::package_name:
+          packages => $::redis::package_name,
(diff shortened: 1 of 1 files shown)


ekohl:
I'm going to close this since I'm working on updates that drop Debian 8.

PR #279, opened 2018-12-05
Make sure mode for /var/run/redis is also set correctly on Ubuntu

bahner:
SSIA mode for /var/run/redis is changed from 2755 to 0755 on each run on Ubuntu


arioch:
Merged, thanks!

PR #282, opened 2018-12-19
use redis 3.2 template for redis 4.x

devfaz:
this would allow usage of protected_mode in 4.x, too


ekohl:
I've merged some serious refactor so this is fixed in another way.

PR #285, opened 2019-02-14
Test bumping versions

cFire:
It works well for me on Debian 9 and puppet 5, but obviously it would be a good idea if more people can test and confirm before merging.


alexjfisher:
This has been fixed in https://github.com/voxpupuli/puppet-redis/pull/296

PR #286, opened 2019-02-25
Bump puppetlabs apt

hp197:
Bump version of puppetlabs apt module.

The old module is broken on Debian >= 9.3.
See this commit: https://github.com/puppetlabs/puppetlabs-apt/commit/c84e842467742f31fba17f243d4c75f9edbc14a9


alexjfisher:
This has been done in https://github.com/voxpupuli/puppet-redis/pull/296 (where the stdlib minimum version was also increased.)

PR #293, opened 2019-03-09
Convert `redisget` to `redis::get` API v4 function

### LINKED ISSUE #291: Convert function to API v4 ruby function
The modern ruby function api is environment isolation safe.

alexjfisher:
Since we'll be dropping puppet 3 support in the next release, we should
move away from using legacy functions that suffer from environment
isolation issues.

See https://puppet.com/docs/puppet/4.10/functions_legacy.html

The newer function API also has built in support for validating function
arguments, declaring some parameters as optional etc.  This allows us to
remove a lot of our own code that was performing this validation before.

Fixes #291


lib/puppet/functions/redis/get.rb
```diff
@@ -0,0 +1,32 @@
+require 'redis'
+# Returns the value of the key being looked up or `undef` if the key does not
+# exist. Takes two arguments with an optional third. The first being a string
+# value of the key to be looked up, the second is the URL to the Redis service
+# and the third optional argument is a default value to be used if the lookup
+# fails.
+#
+# example usage
```
alexjfisher: Unfortunately, https://github.com/puppetlabs/puppet-strings/pull/188 still isn't in a released version of puppet-strings.

alexjfisher: @eputnam Hi! I know you don't work on it anymore, but is there anyone you can suggest pinging to get a new puppet-strings release out?

alexjfisher:
I need to update the README too.

alexjfisher:
README updated

PR #299, opened 2019-03-09
Re-enable and fix acceptance tests. (Don't manage `/var/run/redis` on Debian systems)

alexjfisher:


.sync.yml
```diff
@@ -1,8 +1,15 @@
 ---
+.travis.yml:
+  docker_sets:
+    - set: ubuntu1404-64m
+    - set: ubuntu1604-64m
+    - set: centos6-64m
+    - set: centos7-64m
+    - set: debian8-64m
```
bastelfreak: why is there an `m` at the end of each line?

manifests/config.pp
```diff
@@ -45,28 +45,12 @@
...
-      }
-    }
-
+  unless $facts['osfamily'] == 'Debian' or $service_provider_lookup == 'systemd' {
     file { '/var/run/redis':
       ensure => 'directory',
       owner  => $::redis::config_owner,
-      group  => $var_run_redis_group,
-      mode   => $var_run_redis_mode,
+      group  => $::redis::config_group,
```
bastelfreak: latest modulesync doesn't allow `::` after `$`, so you might want to directly remove it.

ekohl: IMHO not. The module is filled with it and this change keeps it consistent.

alexjfisher: That's only actually for classes.  I think there'll be a separate PR where the variables are sorted out.

spec/acceptance/nodesets/archlinux-2-x64.yml
```diff
@@ -0,0 +1,13 @@
+---
```
ekohl: Would it be an idea to add these nodesets as unmanaged? Longer term I'd like to remove them and if this module isn't using them, let's not pollute its history.

spec/acceptance/suites/default/redis_sentinel_one_node_spec.rb
```diff
@@ -9,6 +9,13 @@
                  'redis'
                end
 
+  # We're testing with `manage_repo` true.  In redis-sentinel 5, the daemon has its own rundir
+  pidfile = if fact('operatingsystem') == 'Ubuntu' && fact('operatingsystemmajrelease') == '16.04'
```
ekohl: Could you implement this in pure puppet below?

.sync.yml
alexjfisher: Role `master` was needed for one of the beaker helpers. `run_task` IIRC.

spec/acceptance/suites/default/redis_sentinel_one_node_spec.rb
alexjfisher: I suppose I may as well.  Ideally, the module itself would set the default correctly, but that's for another day.

spec/acceptance/nodesets/archlinux-2-x64.yml
alexjfisher: Sure.  Can you remind me of the correct syntax for `.sync.yml`?

alexjfisher:
https://github.com/voxpupuli/puppet-redis/pull/299/commits/1e3297e64384339df11a84b385df197a8983423a basically reverts https://github.com/voxpupuli/puppet-redis/pull/173

> rundir directories are 2775

On my system, `/var/run/postgres` was the *only* rundir with that permission.  I went through the distribution redis packages and ppa packages for Debian 7,8,9 and Ubuntu 14.04, 16.04.  They *all* use 0755 (via their init scripts).

spec/acceptance/nodesets/archlinux-2-x64.yml
ekohl: https://github.com/voxpupuli/puppet-windows_env/blob/542466bc58552befc8b209f0b317910c515cc9b3/.sync.yml#L4-L13

spec/acceptance/suites/default/redis_sentinel_one_node_spec.rb
alexjfisher: Done (assuming I've not messed it up).

alexjfisher: I messed it up...

PR #300, opened 2019-03-11
stop managing RUNDIR from the init script

sp-joseluis-ledesma:
Stop managing RUNDIR from the init script on Debian. Init script on Debian runs chmod 755 $RUNDIR, but puppet module may set this permissions to 775, so the redis instance is restarted on every Puppet run.

N/A

### DIFF
--- a/templates/service_templates/redis.Debian.erb
+++ b/templates/service_templates/redis.Debian.erb
@@ -47,10 +47,8 @@ Run_parts () {
 case "$1" in
   start)
   echo -n "Starting $DESC: "
-  mkdir -p $RUNDIR
   touch $PIDFILE
-  chown redis:redis $RUNDIR $PIDFILE
-  chmod 755 $RUNDIR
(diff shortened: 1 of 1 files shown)


alexjfisher:
I think this makes sense since we already the rundir permissions [here](https://github.com/voxpupuli/puppet-redis/blob/f14f07db1be7b1ded2cd7ac233d43bdd5e3681db/manifests/config.pp#L65).

alexjfisher:
Initially, I cherry-picked this into https://github.com/voxpupuli/puppet-redis/pull/299 but their were still some failures.  In the end, I've chosen to not manage `/var/run/redis` on non systemd Debian based systems.

alexjfisher:
https://github.com/voxpupuli/puppet-redis/pull/299 has been merged and contains a fix.
Please let us know if there are still issues.  I wouldn't be surprised if we have to revisit this.

PR #309, opened 2019-04-24
Fix manage_service_file variable

### LINKED ISSUE #308: Fix manage_service_file variable

- Puppet: 4.10.12
- Ruby:
- Distribution: Debian GNU/Linux 9.5 (stretch)
- Module version: v4.0.0

Set 'redis::manage_service_file: true' in hieradata.

## What are you seeing
The block of code that sets up the service file using the template isn't being executed, leaving the file on disk to be the default service file from the redis package install.

That the section of code in manifests/instance.pp should evaluate to true and get executed.
https://github.com/voxpupuli/puppet-redis/blob/950e9e366c071c719a53cc2b03af22e7ff713d7a/manifests/instance.pp#L250-L252

The cause appears to be that 'redis::manage_service_file' isn't defined in the class, though it is defined in params.
https://github.com/voxpupuli/puppet-redis/blob/950e9e366c071c719a53cc2b03af22e7ff713d7a/manifests/params.pp#L38

The config class calls 'redis::instance' trying to access the variable above but seems to always pass false.
https://github.com/voxpupuli/puppet-redis/blob/950e9e366c071c719a53cc2b03af22e7ff713d7a/manifests/config.pp#L32-L40

### LINKED ISSUE #310: Fix systemd service filename

- Puppet: 4.10.12
- Ruby:
- Distribution: Debian GNU/Linux 9.5 (stretch)
- Module version: v4.0.0

Set 'redis::manage_service_file: true' in hieradata.
Set 'redis::service_provider: "systemd"' in hieradata.

## What are you seeing
When the config class creates the default instance it doesn't set the 'redis_server_name' variable, if the instance is set to manage the service file using systemd it creates the file as '/etc/systemd/system/.service'

https://github.com/voxpupuli/puppet-redis/blob/950e9e366c071c719a53cc2b03af22e7ff713d7a/manifests/instance.pp#L219-L226

https://github.com/voxpupuli/puppet-redis/blob/950e9e366c071c719a53cc2b03af22e7ff713d7a/manifests/instance.pp#L256

It appears to be doing as the code says to do, however creating a service file just called '.service' is bad.

CallumBanbery:
The manage_service_file is configured incorrectly and doesn't
allow for overriding the value with hieradata.  Updaing the init.pp
class params allows the file to be passed in.

Fixes #308
Fixes #310

### DIFF
--- a/manifests/instance.pp
+++ b/manifests/instance.pp
@@ -217,6 +217,7 @@ define redis::instance(
 ) {
 
   if $title == 'default' {
+    $redis_server_name    = $::redis::service_name
     $redis_file_name_orig = $config_file_orig
     $redis_file_name      = $config_file
   } else {
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -180,6 +180,7 @@ class redis (
   Stdlib::Filemode $log_dir_mode = $::redis::params::log_dir_mode,
   Stdlib::Absolutepath $log_file = $::redis::params::log_file,
   $log_level                     = $::redis::params::log_level,
+  Boolean $manage_service_file   = $::redis::params::manage_service_file,
   Boolean $manage_package        = $::redis::params::manage_package,
   Boolean $manage_repo           = $::redis::params::manage_repo,
   $masterauth                    = $::redis::params::masterauth,
--- a/spec/classes/redis_spec.rb
+++ b/spec/classes/redis_spec.rb
@@ -1,10 +1,18 @@
 require 'spec_helper'
 
 describe 'redis', type: :class do
+  let(:service_file) { redis_service_file(service_provider: facts[:service_provider]) }
+
   on_supported_os.each do |os, facts|
     context "on #{os}" do
       let(:facts) do
-        facts.merge(redis_server_version: '3.2.3')
+        merged_facts = facts.merge(redis_server_version: '3.2.3')
+
+        if facts[:operatingsystem].casecmp('archlinux') == 0
+          merged_facts = merged_facts.merge(service_provider: 'systemd')
+        end
+
+        merged_facts
       end
 
       let(:package_name) { manifest_vars[:package_name] }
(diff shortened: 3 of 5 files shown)


alexjfisher:
I think this is fixing a bug, but also changing the default behaviour?

Currently `$::redis::manage_service_file` doesn't exist (that's gotta be a bug).
https://github.com/voxpupuli/puppet-redis/blob/950e9e366c071c719a53cc2b03af22e7ff713d7a/manifests/config.pp#L36

ie we end up with
```puppet
redis::instance {'default':
  manage_service_file => undef,
  ...
}
```

This should result in the default value of `true` being used?
https://github.com/voxpupuli/puppet-redis/blob/950e9e366c071c719a53cc2b03af22e7ff713d7a/manifests/instance.pp#L212

With this change, we will now correctly use the value from params.pp, but unfortunately that's a different default.  Does this need to be updated too?

Are you able to add some spec tests to prove/disprove what's actually happening?

CallumBanbery:
@alexjfisher I've added tests for having the `manage_service_file` set to `false` and `true`.
The only thing that's strange is passing in the `service_provider` value.  I tried passing it in the params but it didn't work.
e.g. (Didn't work)
```ruby
let(:params) do
  {
    manage_service_file: false,
    service_provider: 'systemd'
  }
end
```

But passing the value in as a fact and it's fine.
https://github.com/voxpupuli/puppet-redis/blob/30f9b8c924acf35e068588128d66676182f5cf81/spec/classes/redis_spec.rb#L8-L16

I just don't know why that's required as it should really work from `params` because it's a class parameter.

manifests/instance.pp
```diff
@@ -217,6 +217,7 @@
 ) {
 
   if $title == 'default' {
```
dhoppe: Please fix the indentation.

CallumBanbery: @dhoppe Done

dhoppe:
@CallumBanbery Thank you very much for your contribution.

PR #316, opened 2019-07-15
Adding support for more redis cluster params 

dom-nie:
Introducing following parameters for redis cluster:
-cluster-slave-validity-factor 
-cluster-migration-barrier
-cluster-require-full-coverage

None


manifests/init.pp
```diff
@@ -241,6 +250,9 @@
   Boolean $cluster_enabled       = $::redis::params::cluster_enabled,
   $cluster_config_file           = $::redis::params::cluster_config_file,
   $cluster_node_timeout          = $::redis::params::cluster_node_timeout,
+  $cluster_slave_validity_factor = $::redis::params::cluster_slave_validity_factor,
```
bastelfreak: can you please add datatypes to all new parameters and align the `=`?

manifests/instance.pp
```diff
@@ -203,6 +212,9 @@
   Boolean $cluster_enabled       = $::redis::cluster_enabled,
   $cluster_config_file           = $::redis::cluster_config_file,
   $cluster_node_timeout          = $::redis::cluster_node_timeout,
+  $cluster_slave_validity_factor = $::redis::cluster_slave_validity_factor,
```
bastelfreak: same as above

manifests/init.pp
dom-nie: Done

manifests/instance.pp
dom-nie: Done

manifests/init.pp
```diff
@@ -241,6 +250,9 @@
   Boolean $cluster_enabled       = $::redis::params::cluster_enabled,
   $cluster_config_file           = $::redis::params::cluster_config_file,
   $cluster_node_timeout          = $::redis::params::cluster_node_timeout,
+  Integer $cluster_slave_validity_factor = $::redis::params::cluster_slave_validity_factor,
```
bastelfreak: Please align the `=` for all parameters. Can you also define the lowest/highlest allowed value for the integer parameters? At least the lowest value should be enforced. This is probably 0 or 1?

dom-nie: Done.
Both of them have minimum from 0, I have set cluster_migration_barrier default to 1 as per redis docs

PR #318, opened 2019-07-23
Fixing repl_ping_slave_period not being managed with params

dom-nie:
Fixing repl_ping_slave_period not being managed inside template.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -75,7 +75,7 @@
 # @param [String] repl_backlog_size   The replication backlog size
 # @param [String] repl_backlog_ttl   The number of seconds to elapse before freeing backlog buffer
 # @param [String] repl_disable_tcp_nodelay   Enable/disable TCP_NODELAY on the slave socket after SYNC
-# @param [String] repl_ping_slave_period   Slaves send PINGs to server in a predefined interval. It's possible
+# @param [Integer] repl_ping_slave_period   Slaves send PINGs to server in a predefined interval. It's possible
 # @param [String] repl_timeout   Set the replication timeout for:
 # @param [String] requirepass   Require clients to issue AUTH <PASSWORD> before processing any
 #   other commands.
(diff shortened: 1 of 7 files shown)


alexjfisher:
Looks like this was broken in https://github.com/voxpupuli/puppet-redis/commit/7b41977bf328d780c825c603089303e63c0e10e9#diff-fe82d24487e48ff4b7f22a48744defa2R190
I hope just by accident?   ie are you sure this setting is safe in situations you're not using replication?

alexjfisher:
@dom-nie Would you be able to add any tests?

dom-nie:
@alexjfisher 
Test like that is actually there 
https://github.com/voxpupuli/puppet-redis/blob/4921650247d6f549053c1534144ad5412baabf77/spec/classes/redis_spec.rb#L730

I'm not really sure abut case where replication is not used.
I can check it if you think it might have some negative impact.

dom-nie:
> Looks like this was broken in [7b41977#diff-fe82d24487e48ff4b7f22a48744defa2R190](https://github.com/voxpupuli/puppet-redis/commit/7b41977bf328d780c825c603089303e63c0e10e9#diff-fe82d24487e48ff4b7f22a48744defa2R190)
> I hope just by accident? ie are you sure this setting is safe in situations you're not using replication?

Looking at Redis code 
https://github.com/antirez/redis/blob/fde5e737dcaf13f8121ebbc027c617bc593c3bd2/src/server.c#L2351
https://github.com/antirez/redis/blob/fde5e737dcaf13f8121ebbc027c617bc593c3bd2/src/server.h#L109

It's by default always set to 10, so explicitly setting this to default in config should be case.

alexjfisher:
That's too funny.  The test passes because `%r{repl-ping-slave-period.*1}` will happily match `# repl-ping-slave-period 10`

dom-nie:
> That's too funny. The test passes because `%r{repl-ping-slave-period.*1}` will happily match `# repl-ping-slave-period 10`

Fixed

alexjfisher:
@dom-nie Thanks!

PR #239, opened 2017-09-15
Adds support for Redis 4.0

petems:
* Adds new 4.0 config file
* Adds acceptance test for Redis 4.0.1

PR #326, opened 2019-10-23
Drop support for Redis 2, Debian 8 and Ubuntu 14.04; add Debian 9 and Ubuntu 18.04

ekohl:
This is a breaking change. Currently still a WIP. It may need fixes from https://github.com/voxpupuli/puppet-redis/pull/239. It obsoletes https://github.com/voxpupuli/puppet-redis/pull/282 at least.


.sync.yml
```diff
@@ -2,11 +2,12 @@
 .travis.yml:
   secure: "lmmQ5+ueE39cDevK+TjJWgD7b9X6FQDoyL/OTRXzOuCu/96yXQrm/KBq4E2IsFTx91XiqOH6ZnKcgiGwaY54E10ddTvlZhEa8orMmwfkrLAbyh7POB1q3AtQze7RPLvOuHZc6XIRBDiiFKMrzvwdP5OgU9iPEFdSRZll64nw240="
   docker_sets:
-    - set: ubuntu1404-64m
     - set: ubuntu1604-64m
+    - set: ubuntu1804-64m
```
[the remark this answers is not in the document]
ekohl: I considered dropping them since we don't really care about a master or not.

spec/spec_helper.rb
```diff
@@ -94,78 +50,29 @@ def redis_service_file(service_name: redis_service_name, service_provider: nil)
...
-  }
-end
-
-def archlinux_facts
-  {
-    operatingsystem: 'Archlinux',
-    osfamily: 'Archlinux',
-    puppetversion: '4.5.2'
-  }
+add_custom_fact :service_provider, ->(os,facts) do
```
bastelfreak: we should really add this to our modulesync_config or puppetlabs_spec_helper. That's such a commonly used fact.

spec/classes/redis_sentinel_spec.rb
```diff
@@ -1,106 +1,103 @@
...
+        when 'FreeBSD'
+          '/usr/local/etc/redis-sentinel.conf.puppet'
+        when 'RedHat'
+          '/etc/redis-sentinel.conf.puppet'
+        end
+      end
+
+      describe 'without parameters' do
+        let(:expected_content) do
+          <<~EOF
```
ekohl: This is Ruby 2.3 or 2.4 syntax so rubocop breaks on this. Anyone has any idea what the cleanest fix for this is? Should I use the classic `<<EOF` and not indent?

.sync.yml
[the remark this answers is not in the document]
ekohl: Dropping breaks because there is one test that can run on multiple nodes even though we only provision with one.

ekohl:
Not sure about Debian 10 here:
```
  linking shared-object ffi_c.so
  make: /usr/bin/mkdir: Command not found
  make: *** [Makefile:200: .sitearchdir.time] Error 127
```

I guess this is because it's `/bin/mkdir`.

@alexjfisher I think you've done some bolt installation for task testing. Any insight?

spec/spec_helper.rb
ekohl: Possibly. I've been thinking about sets of custom facts per module. Maybe `puppetlabs-stdlib` should even provide such a thing.

ekohl:
I think this is a good first batch. In the next release I'd like to modify service handling because most distros actually come with systemd service files. In the case of Debian/Ubuntu they're way better because they actually provide capability limiting and in more recent versions out of the box multi-instance.

PR #328, opened 2019-11-13
Stricter data types

ekohl:
This sets data types for most parameters that previously had none. It also moves static defaults from params.pp to inline. While this duplicates a few between Redis and the sentinel, the user would already need to know these were linked. The benefit is a much more readable reference documentation.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -15,7 +15,6 @@ _Private Classes_
 
 * `redis::config`: This class provides configuration for Redis.
 * `redis::install`: This class installs the application.
-* `redis::instances` : This class manages multiple instances.
 * `redis::params`: This class provides a number of parameters.
 * `redis::preinstall`: Provides anything required by the install class, such as package
 repositories.
@@ -32,6 +31,7 @@ repositories.
 
 **Data types**
 
+* [`Redis::LogLevel`](#redisloglevel): Specify the server verbosity level.
 * [`Redis::RedisUrl`](#redisredisurl): 
 
 **Tasks**
@@ -83,67 +83,67 @@ The following parameters are available in the `redis` class.
 
 ##### `activerehashing`
 
-Data type: `String`
+Data type: `Boolean`
 
 Enable/disable active rehashing.
 
-Default value: $redis::params::activerehashing
+Default value: `true`
 
 ##### `aof_load_truncated`
 
-Data type: `String`
+Data type: `Boolean`
 
 Enable/disable loading truncated AOF file
 
-Default value: $redis::params::aof_load_truncated
+Default value: `true`
 
 ##### `aof_rewrite_incremental_fsync`
 
-Data type: `String`
+Data type: `Boolean`
 
 Enable/disable fsync for AOF file
 
-Default value: $redis::params::aof_rewrite_incremental_fsync
+Default value: `true`
 
 ##### `appendfilename`
 
-Data type: `String`
+Data type: `String[1]`
 
 The name of the append only file
 
-Default value: $redis::params::appendfilename
+Default value: 'appendonly.aof'
 
 ##### `appendfsync`
 
 Data type: `Enum['no', 'always', 'everysec']`
 
-Adjust fsync mode. Default: `everysec`
+Adjust fsync mode
 
-Default value: $redis::params::appendfsync
+Default value: 'everysec'
 
 ##### `appendonly`
 
-Data type: `String`
+Data type: `Boolean`
 
 Enable/disable appendonly mode.
 
-Default value: $redis::params::appendonly
+Default value: `false`
 
 ##### `auto_aof_rewrite_min_size`
 
-Data type: `String`
+Data type: `String[1]`
 
 Adjust minimum size for auto-aof-rewrite.
 
-Default value: $redis::params::auto_aof_rewrite_min_size
+Default value: '64mb'
 
 ##### `auto_aof_rewrite_percentage`
 
-Data type: `String`
+Data type: `Integer[0]`
 
 Adjust percentatge for auto-aof-rewrite.
 
-Default value: $redis::params::auto_aof_rewrite_percentage
+Default value: 100
 
 ##### `bind`
 
@@ -151,7 +151,7 @@ Data type: `Variant[Stdlib::IP::Address, Array[Stdlib::IP::Address]]`
 
 Configure which IP address(es) to listen on. To bind on all interfaces, use an empty array.
 
-Default value: $redis::params::bind
+Default value: ['127.0.0.1']
 
 ##### `config_dir`
 
@@ -171,7 +171,7 @@ Default value: $redis::params::config_dir_mode
 
 ##### `config_file_orig`
 
-Data type: `String`
+Data type: `Stdlib::Absolutepath`
 
 The location and name of a config file that provides the source
 
@@ -179,7 +179,7 @@ Default value: $redis::params::config_file_orig
 
 ##### `config_file`
 
-Data type: `String`
+Data type: `Stdlib::Absolutepath`
 
 Adjust main configuration file.
 
@@ -191,11 +191,11 @@ Data type: `Stdlib::Filemode`
 
 Adjust permissions for configuration files.
 
-Default value: $redis::params::config_file_mode
+Default value: '0644'
 
 ##### `config_group`
 
-Data type: `String`
+Data type: `String[1]`
 
 Adjust filesystem group for config files.
 
@@ -203,7 +203,7 @@ Default value: $redis::params::config_group
 
 ##### `config_owner`
 
-Data type: `String`
+Data type: `String[1]`
 
 Adjust filesystem owner for config files.
 
@@ -211,19 +211,19 @@ Default value: $redis::params::config_owner
 
 ##### `conf_template`
 
-Data type: `String`
+Data type: `String[1]`
 
 Define which template to use.
 
-Default value: $redis::params::conf_template
+Default value: 'redis/redis.conf.erb'
 
 ##### `daemonize`
 
-Data type: `String`
+Data type: `Boolean`
 
 Have Redis run as a daemon.
 
-Default value: $redis::params::daemonize
+Default value: `true`
 
 ##### `default_install`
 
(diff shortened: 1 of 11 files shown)


REFERENCE.md
```diff
@@ -82,75 +82,75 @@ The following parameters are available in the `redis` class.
 
 ##### `activerehashing`
 
-Data type: `String`
```
alexjfisher: Was this previously documented as a string purely because of explicit puppet-strings tags?  ie it looks like the default was always an actual Boolean.

Dan33l: `String` looks a buggy declarative type in doc where `params.pp` use Boolean :
https://github.com/voxpupuli/puppet-redis/pull/328/files#diff-2f383a197b1a428561b364755701d60eL10

ekohl: Correct. Previously it was set via puppet-strings but it really was a boolean. https://github.com/voxpupuli/puppet-redis/blob/c0d0f76a90abebd197f5410df0ed5764b9a6a0aa/templates/redis.conf.erb#L711 makes this very clear.

ekohl:
> I don't understand why it is flagged `backwards-incompatible`.
> Just someone that use knif like a fork will have incompatibility.

There are changes. Often they're (much) stricter and users can run into this. Where strings used to be accepted, now only integers are. Given the tests are also failing on this indicates I may have been a bit too aggressive in some places.

ekohl:
Note to self: `$hash_max_ziplist_entries` and `$hash_max_ziplist_value` are only relevant on 3.0. Looks like Ubuntu 16.04 is the only platform that ships it out of the box.

igalic:
Can someone explain to me what `'__VALUE__'` is — or was, in this code?

ekohl:
My *guess* is that it was a clear placeholder for easy testing. Now that this is green I plan to go over every config item to see if they're still correct. I already found 2 parameters that no longer exist in 3.2+. That makes me think about dropping Ubuntu 16.04 which is the last that ships Redis 3.0.

igalic:
when will Canonical drop 16.04?

ekohl:
Support stops in April 2021, but I want to add a note that it can work if you install the PPA so you have a version newer than 3.0. Maybe we can set `manage_repo` to true by default on 16.04.

manifests/init.pp
```diff
@@ -19,240 +19,315 @@
...
+  Boolean $activerehashing                                       = true,
+  Boolean $aof_load_truncated                                    = true,
+  Boolean $aof_rewrite_incremental_fsync                         = true,
+  String $appendfilename                                         = 'appendonly.aof',
+  Enum['no', 'always', 'everysec'] $appendfsync                  = 'everysec',
+  Boolean $appendonly                                            = false,
+  String $auto_aof_rewrite_min_size                              = '64mb',
+  Integer[0] $auto_aof_rewrite_percentage                        = 100,
+  Variant[Stdlib::IP::Address, Array[Stdlib::IP::Address]] $bind = ['127.0.0.1'],
+  String $output_buffer_limit_slave                              = '256mb 64mb 60',
```
ghoneycutt: strings should likely be `String[1]` so that you cannot pass `''`. Or `Optional[String[1]]` which would allow for a string or `undef`

manifests/params.pp
```diff
@@ -1,120 +1,26 @@
 # @summary This class provides a number of parameters.
 # @api private
 class redis::params {
```
ghoneycutt: This class should go away completely and be replaced with hiera data in module.

manifests/administration.pp
```diff
@@ -9,16 +9,20 @@
...
 # @author - Peter Souter
+# @see https://redis.io/topics/admin
 #
 class redis::administration(
-  $enable_overcommit_memory = true,
-  $disable_thp              = true,
-  $somaxconn                = '65535',
+  Boolean $enable_overcommit_memory = true,
+  Boolean $disable_thp              = true,
+  String $somaxconn                 = '65535',
```
ghoneycutt: Not familiar with this setting, though if it only accepts integers, then the `Integer` data type might fit better.

manifests/params.pp
ekohl: My motivation to keep it was that this was much easier to understand. I also wanted to limit the changes in this PR to making sure the data types are correct and the REFERENCE.md is useful.

manifests/init.pp
ekohl: Yes, there is room for further tightening. I plan on going over them in another batch.

manifests/administration.pp
ekohl: Good catch. Will check it out.

manifests/params.pp
alexjfisher: There's also more than one class that inherits from params.pp
Getting rid of params.pp in puppet/prometheus left something of a mess IMO.  See https://github.com/voxpupuli/puppet-prometheus/issues/184

manifests/instance.pp
```diff
@@ -8,221 +8,286 @@
...
   Stdlib::Filemode $workdir_mode                                 = $redis::workdir_mode,
-  $zset_max_ziplist_entries                                      = $redis::zset_max_ziplist_entries,
-  $zset_max_ziplist_value                                        = $redis::zset_max_ziplist_value,
+  Integer[0] $zset_max_ziplist_entries                           = $redis::zset_max_ziplist_entries,
+  Integer[0] $zset_max_ziplist_value                             = $redis::zset_max_ziplist_value,
   Boolean $cluster_enabled                                       = $redis::cluster_enabled,
-  $cluster_config_file                                           = $redis::cluster_config_file,
-  $cluster_node_timeout                                          = $redis::cluster_node_timeout,
+  String $cluster_config_file                                    = $redis::cluster_config_file,
+  Integer[0] $cluster_node_timeout                               = $redis::cluster_node_timeout,
```
dom-nie: This value must be 1 or more.

manifests/instance.pp
```diff
@@ -8,221 +8,286 @@
...
+  String[1] $config_group                                        = $redis::config_group,
+  String[1] $config_owner                                        = $redis::config_owner,
+  Boolean $daemonize                                             = $redis::daemonize,
+  Integer[1] $databases                                          = $redis::databases,
+  Variant[String[1], Boolean] $dbfilename                        = $redis::dbfilename,
+  Optional[String] $extra_config_file                            = $redis::extra_config_file,
+  Integer[0] $hash_max_ziplist_entries                           = $redis::hash_max_ziplist_entries,
+  Integer[0] $hash_max_ziplist_value                             = $redis::hash_max_ziplist_value,
+  Integer[0] $hll_sparse_max_bytes                               = $redis::hll_sparse_max_bytes,
+  Integer[0] $hz                                                 = $redis::hz,
```
dom-nie: hz is Integer between 1 and 500.

manifests/instance.pp
```diff
@@ -8,221 +8,286 @@
...
-  $slowlog_max_len                                               = $redis::slowlog_max_len,
-  $stop_writes_on_bgsave_error                                   = $redis::stop_writes_on_bgsave_error,
-  $syslog_enabled                                                = $redis::syslog_enabled,
-  $syslog_facility                                               = $redis::syslog_facility,
-  $tcp_backlog                                                   = $redis::tcp_backlog,
-  $tcp_keepalive                                                 = $redis::tcp_keepalive,
-  $timeout                                                       = $redis::timeout,
-  $unixsocketperm                                                = $redis::unixsocketperm,
-  $ulimit                                                        = $redis::ulimit,
+  Integer[0] $repl_timeout                                       = $redis::repl_timeout,
```
dom-nie: repl-timeout must be 1 or greater

ekohl:
Awesome @dom-nie! Going over the parameters was on my TODO list for today and I highly appreciate it.

manifests/administration.pp
ekohl: Should we change the `if $somaxconn {` to `if $somaxconn > 0 {` to allow users to not manage it?

manifests/sentinel.pp
```diff
@@ -106,34 +106,34 @@
...
-  $failover_timeout       = 180000,
-  $init_script            = $redis::params::sentinel_init_script,
+  Boolean $daemonize = $redis::params::sentinel_daemonize,
+  Integer[0] $down_after = 30000,
+  Integer[0] $failover_timeout = 180000,
+  Optional[Stdlib::Absolutepath] $init_script = $redis::params::sentinel_init_script,
   String[1] $init_template = 'redis/redis-sentinel.init.erb',
-  $log_level              = $redis::params::log_level,
-  $log_file               = $redis::params::log_file,
+  String[1] $log_level = 'notice',
```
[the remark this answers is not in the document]
ekohl: Since this is repeated in 3 files I'll introduce a `Redis::LogLevel` type for this.

manifests/sentinel.pp
```diff
@@ -106,34 +106,34 @@
...
   String[1] $service_name = 'redis-sentinel',
-  $service_ensure         = $redis::params::service_ensure,
-  Boolean $service_enable = $redis::params::service_enable,
-  $service_user           = $redis::params::service_user,
+  Stdlib::Ensure::Service $service_ensure = 'running',
+  Boolean $service_enable = true,
+  String[1] $service_user = 'redis',
   Stdlib::Absolutepath $working_dir = '/tmp',
-  $notification_script    = undef,
+  $notification_script = undef,
```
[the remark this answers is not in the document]
ekohl: The (unit) test uses bar.sh. I don't know if that's actually valid or just an example. The Redis docs aren't explicit about it. Do you know?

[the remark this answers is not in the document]
dom-nie: @ekohl Looking at codebase and docs I couldn't find anything useful as well.
Looking at all examples people were using it online, it was always full path to script.

I think limit it to absolute path is not a bad idea, regardless whether bar.sh would be also accepted.
It will also resolve confusion for other people, and the question where to put this bar.sh to make it work.

What do you think ?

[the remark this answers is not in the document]
ekohl: I agree. It's also easier to go from an absolute path to a string since that's non-breaking.

ekohl:
I think this is now ready for a final review. The `String` entries have been tightened to `String[1]` as well.

Issue #329, opened 2019-11-13
Instance is not read from hiera file


lordbink:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 6.10.1
- Ruby: 2.5.7p206 (default from bin directory inside puppet)
- Distribution: Ubuntu 16.04
- Module version: 4.2.1

## How to reproduce (e.g Puppet code you use)
This works inside nodes.pp:

```
 class { '::redis': }
 redis::instance {'test3':
   port => 1234,
 }
```

This does not work inside the node/hiera file (test-node.eyaml):

```
 redis::service_user: 'redis'
 redis::service_group: 'nogroup'
 redis::instance:
   'test3':
     enabled: true
     port: 1234
```

## What are you seeing
No changes to the end node and it appears to not see the configuration in hiera.

## What behaviour did you expect instead
Update to files inside the /etc/redis directory.

## Output log

> /etc/redis# puppet agent -t --certname test-redis.domain.com --noop
> Info: Using configured environment 'test_branch'
> Info: Retrieving pluginfacts
> Info: Retrieving plugin
> Info: Retrieving locales
> Info: Loading facts
> Info: Applying configuration version '1573688773'
> Notice: Applied catalog in 2.74 seconds

> ll
> total 88
> drwxr-xr-x   6 redis redis  4096 Nov 13 17:44 ./
> drwxr-xr-x 101 root  root   4096 Nov 13 14:56 ../
> -rw-r--r--   1 redis redis 32566 Nov 13 13:59 redis.conf
> -rw-r--r--   1 redis redis 32566 Nov 13 13:59 redis.conf.puppet
> drwxr-xr-x   2 root  root   4096 Nov 13 09:49 redis-server.post-down.d/
> drwxr-xr-x   2 root  root   4096 Nov 13 09:49 redis-server.post-up.d/
> drwxr-xr-x   2 root  root   4096 Nov 13 09:49 redis-server.pre-down.d/
> drwxr-xr-x   2 root  root   4096 Nov 13 09:49 redis-server.pre-up.d/

(nothing applied)

## Any additional information you'd like to impart
Let me know if there is anything I can add to be more helpful.

Dan33l:
Hi @lordbink 

I am not user of this module. But, the resource `redis::instance` is a defined type. And so, it can not be used from hiera data.

[...]

PR #330, opened 2019-11-15
Support instances inside Hiera

lordbink:
https://github.com/voxpupuli/puppet-redis/issues/329

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -259,6 +259,8 @@ class redis (
   contain redis::install
   contain redis::config
   contain redis::service
+  
+  create_resources('::redis::instance', $instances)
 
   Class['redis::preinstall']
   -> Class['redis::install']
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -153,6 +153,7 @@
 # @param [Integer] cluster_migration_barrier    Minimum number of slaves master will remain connected with, for another slave to migrate to a  master which is no longer covered by any slave
 #   Only set if cluster_enabled is true
 #   Default: 1
+# @param [Hash] instances   Iterate through multiple instance configurations
 class redis (
   $activerehashing                                               = $redis::params::activerehashing,
   $aof_load_truncated                                            = $redis::params::aof_load_truncated,
@@ -253,7 +254,7 @@ class redis (
   Integer[0] $cluster_slave_validity_factor                      = $redis::params::cluster_slave_validity_factor,
   Boolean $cluster_require_full_coverage                         = $redis::params::cluster_require_full_coverage,
   Integer[0] $cluster_migration_barrier                          = $redis::params::cluster_migration_barrier,
-  Hash $instances                                                = {},
+  Hash[String[1], Hash] $instances                               = {},
 ) inherits redis::params {
 
   contain redis::preinstall
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -260,8 +260,6 @@ class redis (
   contain redis::config
   contain redis::service
 
-  create_resources('::redis::instance', $::redis::instances)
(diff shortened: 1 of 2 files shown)


manifests/init.pp
```diff
@@ -252,14 +252,18 @@
   $cluster_node_timeout                                          = $redis::params::cluster_node_timeout,
   Integer[0] $cluster_slave_validity_factor                      = $redis::params::cluster_slave_validity_factor,
   Boolean $cluster_require_full_coverage                         = $redis::params::cluster_require_full_coverage,
-  Integer[0] $cluster_migration_barrier                          = $redis::params::cluster_migration_barrier
+  Integer[0] $cluster_migration_barrier                          = $redis::params::cluster_migration_barrier,
+  ## Start hiera lookup
+  $instances                                                     = {},
```
alexjfisher: Could you add a `Hash` data type here?

manifests/init.pp
```diff
@@ -252,14 +252,18 @@
...
 ) inherits redis::params {
 
   contain redis::preinstall
   contain redis::install
   contain redis::config
   contain redis::service
 
+  create_resources('::redis::instance', $instances)
```
alexjfisher: Instead of `create_resources`, could you use iteration instead? (Also drop the leading `::`).
See https://github.com/voxpupuli/puppet-firewalld/blob/master/manifests/init.pp#L103 for an example.

manifests/init.pp
lordbink: Thank you for bringing that up. Done.

manifests/init.pp
lordbink: Again, thank you. Done.

manifests/init.pp
```diff
@@ -252,14 +252,22 @@
   $cluster_node_timeout                                          = $redis::params::cluster_node_timeout,
   Integer[0] $cluster_slave_validity_factor                      = $redis::params::cluster_slave_validity_factor,
   Boolean $cluster_require_full_coverage                         = $redis::params::cluster_require_full_coverage,
-  Integer[0] $cluster_migration_barrier                          = $redis::params::cluster_migration_barrier
+  Integer[0] $cluster_migration_barrier                          = $redis::params::cluster_migration_barrier,
+  ## Start hiera lookup
```
alexjfisher: I think this comment can go. You might be using hiera, but if someone wanted to use `instances` directly in their declaration of the class, that would work equally well.

lordbink: Done.

alexjfisher: Looking good.  Just need to add a puppet strings style comment up [here](https://github.com/voxpupuli/puppet-redis/pull/330/files#diff-60ae41fd0a31977447947f59940ee9a4L155) somewhere, and I think we're good to merge.  (You could regenerate the Reference.md, but not essential as I'm sure someone else will before the next release anyway)

manifests/init.pp
```diff
@@ -252,14 +252,21 @@
   $cluster_node_timeout                                          = $redis::params::cluster_node_timeout,
   Integer[0] $cluster_slave_validity_factor                      = $redis::params::cluster_slave_validity_factor,
   Boolean $cluster_require_full_coverage                         = $redis::params::cluster_require_full_coverage,
-  Integer[0] $cluster_migration_barrier                          = $redis::params::cluster_migration_barrier
+  Integer[0] $cluster_migration_barrier                          = $redis::params::cluster_migration_barrier,
+  Hash $instances                                                = {},
```
bastelfreak: Could you update the datatype?
```suggestion
  Hash[String[1], Hash] $instances                               = {},
```

This should be valid

PR #332, opened 2019-11-27
Allow privileged ports in data types

ekohl:
502ff38c8c035fa143e32dd4ee61b3ec68db461c tightened the data types, but there's no reason to only allow unprivileged ports.


manifests/init.pp
```diff
@@ -283,7 +283,7 @@
   String[1] $package_ensure                                      = 'present',
   String[1] $package_name                                        = $redis::params::package_name,
   Stdlib::Absolutepath $pid_file                                 = $redis::params::pid_file,
-  Stdlib::Port::Unprivileged $port                               = 6379,
+  Stdlib::Port $port                                             = 6379,
```
Dan33l: Can you update `REFERENCE.md` ?

ekohl: And that's why I hate derived files in git. So easy to forget

ekohl:
Updated. Now also updated `instance.pp` and regenerated `REFERENCE.md`.

PR #333, opened 2019-11-27
Allow empty unixsocket(perm)

ekohl:
502ff38c8c035fa143e32dd4ee61b3ec68db461c tightened the data types, but too much. It's valid not to bind on a unix socket and the template allowed this.


REFERENCE.md
```diff
@@ -766,15 +766,15 @@ Default value: 65536
 
 ##### `unixsocket`
 
-Data type: `Stdlib::Absolutepath`
+Data type: `Variant[Stdlib::Absolutepath, Enum['']]`
```
Dan33l: With this data type the PR is backward-incompatible because it is not possible to use `false` to disable unix socket.

manifests/init.pp
```diff
@@ -317,8 +317,8 @@
   Integer[0] $tcp_backlog                                        = 511,
   Integer[0] $tcp_keepalive                                      = 0,
   Integer[0] $timeout                                            = 0,
-  Stdlib::Absolutepath $unixsocket                               = '/var/run/redis/redis.sock',
-  Stdlib::Filemode $unixsocketperm                               = '0755',
+  Variant[Stdlib::Absolutepath, Enum['']] $unixsocket            = '/var/run/redis/redis.sock',
+  Variant[Stdlib::Filemode, Enum['']] $unixsocketperm            = '0755',
```
Dan33l: Is it really needed ?

templates/redis.conf.erb
```diff
@@ -66,8 +66,8 @@ bind <%= @bind_arr.join(' ') %>
 # incoming connections. There is no default, so Redis will not listen
 # on a unix socket when not specified.
 #
-<% if @unixsocket %>unixsocket <%= @unixsocket %><% end %>
-<% if @unixsocketperm %>unixsocketperm <%= @unixsocketperm %><% end %>
+<% unless @unixsocket.empty? %>unixsocket <%= @unixsocket %><% end %>
+<% unless @unixsocketperm.empty? %>unixsocketperm <%= @unixsocketperm %><% end %>
```
Dan33l: The `unixsocketperm` should pop in the file only if `unixsocket` is not empty.
By this way, the user of the module does have to take care of `unixsocketperm` when `unixsocket` is disabled.

REFERENCE.md
ekohl: I marked the original PR as backwards-incompatible. This particular PR is backwards compatible.

templates/redis.conf.erb
ekohl: I like that idea.

manifests/init.pp
ekohl: Previously the template allowed it.

ekohl:
How about this version?

manifests/init.pp
Dan33l: I see only one use case , let the default umask manage the perm of the file.
Bah ... not sure it is useful, but why not.

ekohl:
@amoralej mind having a look?

PR #334, opened 2019-11-27
Add SCL support

ekohl:
On Red Hat the software collections allow for multiple versions. On RHEL this is also supported for customers where EPEL is unsupported.

To implement this, the globals pattern is implemented giving users an easy way to select the SCL at a global level. Acceptance tests are added to ensure it works.

For convenience, it can ensure centos-release-scl-rh is installed on CentOS when manage_repo is true.

This is a draft because I'm still working on verification. The acceptance tests are also expected to fail since it'll attempt to bind on the same port without stopping the old service.


ekohl:
I believe this is ready for review now. To make this easier I've split it in several commits.

PR #335, opened 2019-11-28
Clean up preinstall handling and use modern facts

ekohl:


manifests/preinstall.pp
```diff
@@ -3,41 +3,22 @@
 # @api private
 class redis::preinstall {
   if $redis::manage_repo {
-    case $::operatingsystem {
-      'RedHat', 'CentOS', 'Scientific', 'OEL', 'Amazon': {
+    if $facts['os']['family'] == 'RedHat' {
```
runejuhl: I'd prefer that the `case` statement is kept, so that it can have a `default` clause that fails instead of silently ignoring it.

ekohl: Hmm, I'll split it up in a different PR.

PR #336, opened 2019-11-29
Closer match parameter to OS defaults

ekohl:
This should match what's in EPEL, Debian and Ubuntu.

It's split off from https://github.com/voxpupuli/puppet-redis/pull/334 to make sure the tests pass first. It's also more of a bugfix than a feature.


manifests/instance.pp
```diff
@@ -209,14 +207,12 @@
...
   String[1] $conf_template                                       = $redis::conf_template,
-  Stdlib::Absolutepath $config_dir                               = $redis::config_dir,
-  Stdlib::Filemode $config_dir_mode                              = $redis::config_dir_mode,
   Stdlib::Absolutepath $config_file                              = $redis::config_file,
   Stdlib::Filemode $config_file_mode                             = $redis::config_file_mode,
   Stdlib::Absolutepath $config_file_orig                         = $redis::config_file_orig,
   String[1] $config_group                                        = $redis::config_group,
   String[1] $config_owner                                        = $redis::config_owner,
-  Boolean $daemonize                                             = $redis::daemonize,
+  Boolean $daemonize                                             = true,
```
ekohl: The service scripts we have only work when daemonized. In config.pp we override it for the default to allow the unmanaged service on RH-based systems to work undaemonized.

Issue #322, opened 2019-08-26
Support for Debian 10?


hpkunz:
Hi,

I am using 

- Puppet: 6.8.0
- Ruby:
- Distribution: Debian 10 (buster)
- Module version: 4.2.0

I just installed redis/sentinel on a Debian buster machine and integrated it into an existing group of existing redis servers (replication). I observed the following:

1. redis-server worked as expected
2. starting the redis-sentinel always resulted in timeouts (although the service itself seemed to be running)

I realized that the puppet/redis module adjust the /etc/init.d/redis-sentinel startup script, but not the /etc/systemd/system/sentinel.service file. Note that in debian (since stretch I think) init.d and systemd scripts can co-exist. it seems that the systemd script takes precedence.

I could get around problem 2. (see above) by setting

class { 'redis::sentinel':
    [...]
    pid_file => '/var/run/sentinel/redis-sentinel.pid'
  }

this is the location, where the systemd startup script checks for the pid file.

I believe the proper way to go would be to adjust that module that it adjusts the systemd script.

I you give me a few instructions how I could do that I am willing to give this a shot.

Best,
Hp

ekohl:
fce8da9c7b977db9f87e236cf63c2cb97abe8f93 has updated the pid file to that location on Debian. I had some issues with the bolt gem on Debian for the acceptance tests so it's not officially marked as supported. I'll gladly merge patches to add that support.

basti-nis:
I got the same problem, but your new pid file path won't work for me.

> Distributor ID: Debian
Description:    Debian GNU/Linux 10 (buster)
Release:        10
Codename:       buster
Debian 10.2


> redis-sentinel:
  Installed: 5:5.0.3-4+deb10u1
  Candidate: 5:5.0.3-4+deb10u1
  Version table:
 *** 5:5.0.3-4+deb10u1 500


The pid file path for me is as follows:
`/run/sentinel/redis-sentinel.pid`

After i changed the parameter to:

`pid_file => '/run/sentinel/redis-sentinel.pid'`

It worked like a charm.


Seems there is no consistency.


Which versions do you use?

PR #338, opened 2019-12-05
Debian10 fix

gery6:
Fix $sentinel_pid_file  for Debian.
Systemd will lookup for the pidfile in /var/run/sentinel/redis-sentinel.pid

Debian10 default path for pidfile of sentinel will -> /var/run/sentinel/redis-sentinel.pid

#322

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -23,7 +23,14 @@ class redis::params inherits redis::globals {
       $sentinel_daemonize        = true
       $sentinel_init_script      = '/etc/init.d/redis-sentinel'
       $sentinel_package_name     = 'redis-sentinel'
-      $sentinel_pid_file         = '/var/run/sentinel/redis-sentinel.pid'
+      
+     if versioncmp($facts['os']['release']['full'], '10.0.0') >= 0 {
+       $sentinel_pid_file         = '/var/run/sentinel/redis-sentinel.pid'
+      } else {
+       $sentinel_pid_file         = '/var/run/redis/redis-sentinel.pid'
+      }
+
+      
       $sentinel_log_file         = '/var/log/redis/redis-sentinel.log'
       $sentinel_working_dir      = '/var/lib/redis'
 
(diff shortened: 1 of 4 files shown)


gery6:
@dhoppe: You're right. Debian Stretch has another path. 
I will spend some work on it, soon, when the time will allows it.

Thank you.

basti-nis:
I got the same problem, but your new pid file path won't work for me.

> Distributor ID: Debian
Description:    Debian GNU/Linux 10 (buster)
Release:        10
Codename:       buster
Debian 10.2


> redis-sentinel:
  Installed: 5:5.0.3-4+deb10u1
  Candidate: 5:5.0.3-4+deb10u1
  Version table:
 *** 5:5.0.3-4+deb10u1 500


The pid file path for me is as follows:
`/run/sentinel/redis-sentinel.pid`

After i changed the parameter to:

`pid_file => '/run/sentinel/redis-sentinel.pid'`

It worked like a charm.


Seems there is no consistency.


Which versions do you use?

manifests/params.pp
```diff
@@ -23,7 +23,13 @@
       $sentinel_daemonize        = true
       $sentinel_init_script      = '/etc/init.d/redis-sentinel'
       $sentinel_package_name     = 'redis-sentinel'
+
+      if versioncmp($facts['os']['release']['full'], '10.0.0') >= 0 {
+      $sentinel_pid_file         = '/var/run/sentinel/redis-sentinel.pid'
```
ekohl:
```suggestion
      $sentinel_pid_file         = '/run/sentinel/redis-sentinel.pid'
```

ekohl: This was marked as resolved, but I don't see the actual change so I've marked it as unresolved. Mind having a look?

spec/classes/redis_sentinel_spec.rb
```diff
@@ -10,6 +10,8 @@
           '/etc/redis/redis-sentinel.conf.puppet'
         when 'Debian'
           '/etc/redis/redis-sentinel.conf.puppet'
+        when 'Ubuntu'
```
ekohl: There is no Ubuntu OS family since Ubuntu is part of the Debian OS family.

PR #340, opened 2019-12-23
modulesync 2.10.0

bastelfreak:
modulesync 2.10.0


ghoneycutt:
Re-kicked a failing travis job and now all is well. Merging.

PR #342, opened 2020-01-07
fix service template

basti-nis:
If you don't change the default socket pid file ($pid_file) it's necessary to create the redis-folder via systemd-unit-file otherwise you will face the problem "Opening Unix socket: bind: No such file or directory" on service start.

### DIFF
--- a/templates/service_templates/redis.service.erb
+++ b/templates/service_templates/redis.service.erb
@@ -3,6 +3,8 @@ Description=Advanced key-value store for <%= @title %>
 After=network.target
 
 [Service]
+RuntimeDirectory=redis
+RuntimeDirectoryMode=0755
 Type=forking
 ExecStart=/usr/bin/redis-server <%= @redis_file_name %>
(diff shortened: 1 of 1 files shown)


templates/service_templates/redis.service.erb
```diff
@@ -3,6 +3,8 @@ Description=Advanced key-value store for <%= @title %>
 After=network.target
 
 [Service]
+RuntimeDirectory=redis
+RuntimeDirectoryMode=0755
```
dhoppe: Could you please test this with ```RuntimeDirectoryMode=2755```? I have seen this at Debian and Ubuntu.

basti-nis: Ok!
I'll give you the test results in a few minutes.

basti-nis: Testing finished with Debian 10.2 and Redis package version 5.0.3-4+deb10u1

Work's like a charm!

PR #343, opened 2020-01-21
Finishing touches for multi-instance support

fraenki:
This is essentially a stripped-down re-submission of #230. The original author of #230 seems to have abandoned his efforts (the repository is gone).

This PR contains the following enhancements:

* Add an example for multiple instances to the README
* Fix a dependency cycle by using camptocamp/systemd instead of `Exec` to reload systemd
* Updates to the systemd template
* Update tests accordingly

However, this re-submission does not include the `redis-shutdown.erb` template that was present in the original PR, because some users noted that this is not required. I can confirm this from my own testing.

Tested on CentOS 7.7 with Puppet 6.11 and Redis 5.

FWIW, this is the dependency cycle that may occur without the proposed modifications when using multiple instances:

```
Error: Found 1 dependency cycle:
(Exec[systemd-reload-redis] => Class[Redis] => Redis::Instance[6379] => File[/etc/systemd/system/redis-server-6379.service] => Exec[systemd-reload-redis])
```

n/a

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -11,3 +11,6 @@ fixtures:
     yumrepo_core:
       repo: "https://github.com/puppetlabs/puppetlabs-yumrepo_core"
       puppet_version: ">= 6.0.0"
+    systemd:
+      repo: 'https://github.com/camptocamp/puppet-systemd.git'
+      puppet_version: "< 6.1.0"
--- a/README.md
+++ b/README.md
@@ -64,6 +64,33 @@ class { '::redis':
 }
 ```
 
+### Multiple instances
+
+
+```puppet
+$listening_ports = [6379,6380,6381,6382]
+
+class { '::redis':
+  default_install => false,
+  service_enable  => false,
+  service_ensure  => 'stopped',
+}
+
+$listening_ports.each |$port| {
+  $port_string = sprintf('%d',$port)
+  redis::instance { $port_string:
+    service_enable => true,
+    service_ensure => 'running',
+    port           => $port,
+    bind           => $facts['networking']['ip'],
+    dbfilename     => "${port}-dump.rdb",
+    appendfilename => "${port}-appendonly.aof",
+    appendfsync    => 'always',
+    require        => Class['Redis'],
+  }
+}
+```
+
 ### Manage repositories
 
 Disabled by default but if you really want the module to manage the required
@@ -103,6 +130,10 @@ class { '::redis::sentinel':
   failover_timeout => 30000,
 }
 ```
+
+### Soft dependency
+
+This module requires `camptocamp/systemd` on Puppet versions older than 6.1.0.
 
 ## `redis::get()` function
 
--- a/manifests/instance.pp
+++ b/manifests/instance.pp
@@ -286,7 +286,6 @@ define redis::instance (
   Variant[Stdlib::Absolutepath, Enum['']] $unixsocket            = "/var/run/redis/redis-server-${name}.sock",
   Stdlib::Absolutepath $workdir                                  = "${redis::workdir}/redis-server-${name}",
 ) {
-
   if $title == 'default' {
     $redis_file_name_orig = $config_file_orig
     $redis_file_name      = $config_file
@@ -330,7 +329,13 @@ define redis::instance (
         mode    => '0644',
         content => template('redis/service_templates/redis.service.erb'),
       }
-      ~> Exec['systemd-reload-redis']
+
+      # Only necessary for Puppet < 6.1.0,
+      # See https://github.com/puppetlabs/puppet/commit/f8d5c60ddb130c6429ff12736bfdb4ae669a9fd4
+      if versioncmp($facts['puppetversion'],'6.1.0') < 0 {
+        include systemd::systemctl::daemon_reload
+        File["/etc/systemd/system/${service_name}.service"] ~> Class['systemd::systemctl::daemon_reload']
+      }
 
       if $title != 'default' {
         service { $service_name:
--- a/manifests/ulimit.pp
+++ b/manifests/ulimit.pp
@@ -49,9 +49,12 @@ class redis::ulimit {
         "defnode nofile Service/LimitNOFILE \"\"",
         "set \$nofile/value \"${redis::ulimit}\"",
       ],
-      notify  => [
-        Exec['systemd-reload-redis'],
-      ],
+    }
+    # Only necessary for Puppet < 6.1.0,
+    # See https://github.com/puppetlabs/puppet/commit/f8d5c60ddb130c6429ff12736bfdb4ae669a9fd4
+    if versioncmp($facts['puppetversion'],'6.1.0') < 0 {
+      include systemd::systemctl::daemon_reload
+      Augeas['Systemd redis ulimit'] ~> Class['systemd::systemctl::daemon_reload']
     }
   } else {
     case $facts['os']['family'] {
--- a/spec/acceptance/suites/default/redis_multi_instances_one_host_spec.rb
+++ b/spec/acceptance/suites/default/redis_multi_instances_one_host_spec.rb
@@ -1,28 +1,40 @@
 require 'spec_helper_acceptance'
 
-describe 'redis::instance' do
+describe 'redis::instance example' do
+  instances = [6379, 6380, 6381, 6382]
   case fact('osfamily')
   when 'Debian'
     config_path = '/etc/redis'
     redis_name = 'redis-server'
   else
-    redis_name = 'redis'
     config_path = '/etc'
+    redis_name = 'redis'
   end
 
   it 'runs successfully' do
     pp = <<-EOS
-    class { 'redis':
+    $listening_ports = #{instances}
+
+    class { '::redis':
       default_install => false,
+      service_enable  => false,
+      service_ensure  => 'stopped',
     }
 
-    redis::instance {'redis1':
-      port => 7777,
+    $listening_ports.each |$port| {
+      $port_string = sprintf('%d',$port)
+      redis::instance { $port_string:
+        service_enable => true,
+        service_ensure => 'running',
+        port           => $port,
+        bind           => $facts['networking']['ip'],
+        dbfilename     => "${port}-dump.rdb",
+        appendfilename => "${port}-appendonly.aof",
+        appendfsync    => 'always',
+        require        => Class['Redis'],
+      }
     }
 
-    redis::instance {'redis2':
-      port => 8888,
-    }
     EOS
 
     # Apply twice to ensure no errors the second time.
@@ -34,29 +46,32 @@ describe 'redis::instance' do
     it { is_expected.to be_installed }
   end
 
-  describe service('redis-server-redis1') do
-    it { is_expected.to be_running }
+  describe service(redis_name) do
+    it { is_expected.not_to be_enabled }
+    it { is_expected.not_to be_running }
   end
 
-  describe service('redis-server-redis2') do
-    it { is_expected.to be_running }
-  end
-
-  describe file("#{config_path}/redis-server-redis1.conf") do
-    its(:content) { is_expected.to match %r{port 7777} }
-  end
-
-  describe file("#{config_path}/redis-server-redis2.conf") do
-    its(:content) { is_expected.to match %r{port 8888} }
-  end
-
-  context 'redis should respond to ping command' do
-    describe command('redis-cli -h 127.0.0.1 -p 7777 ping') do
-      its(:stdout) { is_expected.to match %r{PONG} }
+  instances.each do |instance|
+    describe file("/etc/systemd/system/redis-server-#{instance}.service"), if: (fact('service_provider') == 'systemd') do
+      its(:content) { is_expected.to match %r{redis-server-#{instance}.conf} }
     end
 
-    describe command('redis-cli -h 127.0.0.1 -p 8888 ping') do
-      its(:stdout) { is_expected.to match %r{PONG} }
+    describe service("redis-server-#{instance}") do
+      it { is_expected.to be_enabled }
+    end
+
+    describe service("redis-server-#{instance}") do
+      it { is_expected.to be_running }
+    end
+
+    describe file("#{config_path}/redis-server-#{instance}.conf") do
+      its(:content) { is_expected.to match %r{port #{instance}} }
+    end
+
+    context "redis instance #{instance} should respond to ping command" do
+      describe command("redis-cli -h #{fact('networking.ip')} -p #{instance} ping") do
+        its(:stdout) { is_expected.to match %r{PONG} }
+      end
     end
   end
 end
--- a/spec/classes/redis_spec.rb
+++ b/spec/classes/redis_spec.rb
@@ -4,11 +4,20 @@ describe 'redis' do
   let(:service_file) { redis_service_file(service_provider: facts['service_provider']) }
   let(:package_name) { manifest_vars[:package_name] }
   let(:service_name) { manifest_vars[:service_name] }
+  let(:config_file) { manifest_vars[:config_file] }
   let(:config_file_orig) { manifest_vars[:config_file_orig] }
 
   on_supported_os.each do |os, facts|
     context "on #{os}" do
       let(:facts) { facts }
+
+      if facts[:operatingsystem] == 'Ubuntu' && facts[:operatingsystemmajrelease] == '16.04'
+        let(:systemd) { '' }
+        let(:servicetype) { 'forking' }
+      else
+        let(:systemd) { ' --supervised systemd' }
+        let(:servicetype) { 'notify' }
+      end
 
       describe 'without parameters' do
         it { is_expected.to compile.with_all_deps }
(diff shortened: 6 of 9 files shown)


ghoneycutt:
@fraenki Great work! Could you please add some acceptance tests that show this is working with multiple instances? That would make this easy to merge and greatly help ensure that the functionality remains in tact.

fraenki:
@ghoneycutt Sure, I'll add some acceptance tests :)

fraenki:
@ghoneycutt So I took a look at the acceptance tests and the existing tests ([this](https://github.com/voxpupuli/puppet-redis/blob/master/spec/acceptance/suites/default/redis_multi_instances_one_host_spec.rb)) already have good coverage for multi instance support and they [ran successfully](https://travis-ci.org/voxpupuli/puppet-redis/jobs/639963488) for this PR. Is there anything else I should add?

README.md
```diff
@@ -64,6 +64,33 @@ class { '::redis':
 }
 ```
 
+### Multiple instances
+
+
+```puppet
```
ghoneycutt: This code would be ideal to copy/paste into an acceptance test and then run a command to test that redis can connect to each of the ports or similar to show that it works.

templates/service_templates/redis.service.erb
```diff
@@ -1,15 +1,19 @@
...
 Type=forking
-ExecStart=/usr/bin/redis-server <%= @redis_file_name %>
+ExecStart=/usr/bin/redis-server <%= @redis_file_name %> --supervised systemd
 ExecStop=/usr/bin/redis-server -p <%= @port %> shutdown
-TimeoutStopSec=0
 Restart=always
 User=<%= @service_user %>
 Group=<%= @service_user %>
+RuntimeDirectory=redis
+RuntimeDirectoryMode=0755
```
ghoneycutt: A lot is changing here. Could you add a spec test that shows the contents of this file is being rendered appropriately.

Here's an example - https://github.com/ghoneycutt/puppet-module-bind/blob/master/spec/defines/key_spec.rb#L26-L50

.fixtures.yml
```diff
@@ -11,3 +11,4 @@ fixtures:
     yumrepo_core:
       repo: "https://github.com/puppetlabs/puppetlabs-yumrepo_core"
       puppet_version: ">= 6.0.0"
+    systemd: 'https://github.com/camptocamp/puppet-systemd.git'
```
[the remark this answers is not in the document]
fraenki: It is required to solve a dependency cycle.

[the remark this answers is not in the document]
bastelfreak: I think it's good to keep using the camptocamp module until we can get rid of Puppet 5 by the end of the year. We currently use their exec resource in most of the modules.

README.md
fraenki: I've modified the original multi-instance acceptance test to use the example code from the README.

README.md
```diff
@@ -64,6 +64,33 @@ class { '::redis':
...
+$listening_ports = [6379,6380,6381,6382]
+
+class { '::redis':
```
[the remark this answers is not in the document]
fraenki: Currently all examples in the README use this notation. Should I update all examples?

manifests/instance.pp
```diff
@@ -330,7 +331,7 @@
         mode    => '0644',
         content => template('redis/service_templates/redis.service.erb'),
       }
-      ~> Exec['systemd-reload-redis']
+      ~> Class['systemd::systemctl::daemon_reload']
```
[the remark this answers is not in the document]
fraenki: Fixed.

templates/service_templates/redis.service.erb
fraenki: Added a spec test for the systemd service file.

fraenki:
@ghoneycutt I've addressed all issues:

* Update acceptance tests to use new example code from README
* Re-enable multi-instance acceptance tests on Debian (because they are working fine)
* Add spec test for systemd service file
* Add condition to reload systemd only on Puppet < 6.1.0
* Rebase against current master

templates/service_templates/redis.service.erb
```diff
@@ -1,17 +1,25 @@
 [Unit]
-Description=Advanced key-value store for <%= @title %>
+Description=Redis Advanced key-value store for instance <%= @title %>
```
bastelfreak: since there are so many changes in this template it would be awesome if it would be an epp template and not erb. Feel to convert it if you like it. We only require epp for completely new templates.

manifests/instance.pp
```diff
@@ -286,6 +286,7 @@
   Variant[Stdlib::Absolutepath, Enum['']] $unixsocket            = "/var/run/redis/redis-server-${name}.sock",
   Stdlib::Absolutepath $workdir                                  = "${redis::workdir}/redis-server-${name}",
 ) {
+  include systemd::systemctl::daemon_reload
```
ekohl: Perhaps this should be inside the if condition below?

spec/acceptance/suites/default/redis_multi_instances_one_host_spec.rb
```diff
@@ -1,28 +1,44 @@
...
-    redis_name = 'redis'
     config_path = '/etc'
+    redis_name = 'redis'
   end
 
   it 'runs successfully' do
     pp = <<-EOS
-    class { 'redis':
-      default_install => false,
+    Exec {
```
ekohl: Is this exec needed? I'm not sure there's a lot of execs left in this module and I got it to work without.

spec/acceptance/suites/default/redis_multi_instances_one_host_spec.rb
```diff
@@ -1,28 +1,44 @@
...
+      path => [ '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin', ]
     }
 
-    redis::instance {'redis1':
-      port => 7777,
+    $listening_ports = #{instances}
+
+    class { '::redis':
+      default_install => false,
+      service_enable  => false,
```
ekohl: If you disable this, could you add a test `describe service(redis_name) { it { is_expected.not_to be_enabled } }` and the same for the line below?

spec/acceptance/suites/default/redis_multi_instances_one_host_spec.rb
```diff
@@ -34,29 +50,23 @@ class { 'redis':
...
+  instances.each do |instance|
+    describe file("/etc/systemd/system/redis-server-#{instance}.service"), if: (fact('service_provider') == 'systemd') do
+      its(:content) { is_expected.to match %r{redis-server-#{instance}.conf} }
+    end
 
-  describe file("#{config_path}/redis-server-redis2.conf") do
-    its(:content) { is_expected.to match %r{port 8888} }
-  end
+    describe service("redis-server-#{instance}") do
+      it { is_expected.to be_running }
```
ekohl: Mind adding `be_enabled` as well while you're at it?

spec/acceptance/suites/default/redis_multi_instances_one_host_spec.rb
```diff
@@ -34,29 +50,23 @@ class { 'redis':
...
-    describe command('redis-cli -h 127.0.0.1 -p 7777 ping') do
-      its(:stdout) { is_expected.to match %r{PONG} }
+    describe file("#{config_path}/redis-server-#{instance}.conf") do
+      its(:content) { is_expected.to match %r{port #{instance}} }
     end
 
-    describe command('redis-cli -h 127.0.0.1 -p 8888 ping') do
-      its(:stdout) { is_expected.to match %r{PONG} }
+    context "redis instance #{instance} should respond to ping command" do
+      describe command("redis-cli -h #{fact('networking.ip')} -p #{instance} ping") do
```
ekohl: Can this use localhost instead of the IP?

spec/classes/redis_spec.rb
```diff
@@ -1329,6 +1336,34 @@ class { 'redis::globals':
         end
 
         it { is_expected.to contain_file(service_file) }
+
+        it do
+          content = <<-END.gsub(%r{^\s+\|}, '')
```
ekohl: I'm not sure we can use `<<~END` with our rubocop config, but that's the modern Ruby way.

templates/service_templates/redis.service.erb
```diff
@@ -1,17 +1,24 @@
...
 After=network.target
+After=network-online.target
+Wants=network-online.target
 
 [Service]
 RuntimeDirectory=redis
 RuntimeDirectoryMode=2755
 Type=forking
```
ekohl: If it's supervised, I think the type should also be `notify` for optimal integration.

spec/acceptance/suites/default/redis_multi_instances_one_host_spec.rb
fraenki: No, because we are testing the example from the README, which uses `bind => $facts['networking']['ip']`. Hence it would not work with localhost.

spec/classes/redis_spec.rb
fraenki: The tests do not complain, so I guess it works? :)

spec/acceptance/suites/default/redis_multi_instances_one_host_spec.rb
fraenki: Sure, will do.

spec/acceptance/suites/default/redis_multi_instances_one_host_spec.rb
fraenki: Sure, will do.

spec/acceptance/suites/default/redis_multi_instances_one_host_spec.rb
fraenki: This looks like a copy'n'paste leftover, I'll remove it.

manifests/instance.pp
fraenki: Sure, will do.

templates/service_templates/redis.service.erb
fraenki: Sure, will do.

spec/acceptance/suites/default/redis_multi_instances_one_host_spec.rb
```diff
@@ -34,29 +46,31 @@ class { 'redis':
     it { is_expected.to be_installed }
   end
 
-  describe service('redis-server-redis1') do
-    it { is_expected.to be_running }
+  describe service(redis_name) do
```
ekohl: Could you add `not_to be_running` as well?

spec/classes/redis_spec.rb
```diff
@@ -1329,6 +1338,34 @@ class { 'redis::globals':
...
+            |Restart=always
+            |User=redis
+            |Group=redis
+            |LimitNOFILE=65536
+            |
+            |[Install]
+            |WantedBy=multi-user.target
+          END
+
+          if facts['service_provider'] == 'systemd'
```
ekohl: I always like to add an `else is_expected.not_to contain_file(service_file)` to these kind of situations.

manifests/ulimit.pp
```diff
@@ -50,7 +51,7 @@
         "set \$nofile/value \"${redis::ulimit}\"",
       ],
       notify  => [
-        Exec['systemd-reload-redis'],
+        Class['systemd::systemctl::daemon_reload'],
```
ekohl: Should this notify also be hidden behind an if condition?

fraenki: OK, will add a condition.

fraenki:
> On a related note, I'm sure you've noticed that more recent Debian (Buster) and Ubuntu (Bionic) versions have started to ship the `redis(-server)@` service. I'd like to use those since they also contain various security hardening features. Not a blocker to this, but have you had a look at this and what are your thoughts?

Not at all, actually I'm not doing anything with Debian or Ubuntu.
So I'd prefer if this could be integrated with another PR. :)

ekohl:
> Not at all, actually I'm not doing anything with Debian or Ubuntu.
> So I'd prefer if this could be integrated with another PR. :)

Absolutely. However, I do wonder if we could deploy a `redis(-server)@` unit file on platforms that are missing it and utilize that. Since you're not a Debian/Ubuntu user, how would you feel about that?

spec/acceptance/suites/default/redis_multi_instances_one_host_spec.rb
fraenki: Sure, will do.

spec/classes/redis_spec.rb
fraenki: Sure, will add it.

fraenki: Actually, `not_to` would be wrong here. This test also works for non-systemd systems, but apparently with different content. I've added a small test for non-systemd systems instead.

manifests/ulimit.pp
```diff
@@ -17,6 +17,13 @@
 class redis::ulimit {
   assert_private('The redis::ulimit class is only to be called from the redis::config class')
 
+  # Only necessary for Puppet < 6.1.0,
```
ekohl: This should be inside the `if $service_provider_lookup == 'systemd' {`. Otherwise it's included on non-systemd distros. I'd also write `Augeas['Systemd redis ulimit'] ~> Class['systemd::systemctl::daemon_reload']` instead of using `$_notify`

metadata.json
```diff
@@ -27,6 +27,10 @@
     {
       "name": "herculesteam/augeasproviders_core",
       "version_requirement": ">= 2.1.0 < 3.0.0"
+    },
+    {
+      "name": "camptocamp/systemd",
```
ekohl: Given this is only needed on systems that run systemd and Puppet < 6.1.0, should we make this a soft dependency documented in README?

manifests/ulimit.pp
fraenki: Sure, will also add this condition to the test so it does not fail anymore.

metadata.json
fraenki: Removed dependency and added a section to the README.

README.md
```diff
@@ -64,6 +64,33 @@ class { '::redis':
...
+}
+
+$listening_ports.each |$port| {
+  $port_string = sprintf('%d',$port)
+  redis::instance { $port_string:
+    service_enable => true,
+    service_ensure => 'running',
+    port           => $port,
+    bind           => $facts['networking']['ip'],
+    dbfilename     => "${port}-dump.rdb",
```
ekohl: This makes me wonder if we should have some code in `instance.pp` that does this and appendfilename so they're unique. Defaulting to the same file for multiple instances sounds like a bad idea. We can pass the default in `config.pp` as we do for other `*file` parameters.

.fixtures.yml
```diff
@@ -11,3 +11,6 @@ fixtures:
     yumrepo_core:
       repo: "https://github.com/puppetlabs/puppetlabs-yumrepo_core"
       puppet_version: ">= 6.0.0"
+    systemd:
+      repo: 'https://github.com/camptocamp/puppet-systemd.git'
+      puppet_version: "< 6.1.0"
```
ekohl: I believe this is the first use I've seen that isn't `>= 6.0.0` but nice!

README.md
fraenki: While I agree, I think having `$port` as the one and only unique identifier for Redis instances is sufficient. Adding `$bind` to the mix by introducing it as unique identifier could make things pretty complicated. I think this would only add choice (which is usually a good thing), but without real benefits.

fraenki:
> Absolutely. However, I do wonder if we could deploy a redis(-server)@ unit file on platforms that are missing it and utilize that. Since you're not a Debian/Ubuntu user, how would you feel about that?

I took a look at this. It sure is an improvement. I took note and will implement it in a follow-up PR, because this one is already too big and I don't want to see all checks break again. :) (The follow-up PR would also be a good opportunity to convert the service template to EPP, as suggested by @bastelfreak.)

Now that all checks are working... is this ready for merge? :)

spec/classes/redis_spec.rb
```diff
@@ -99,6 +108,7 @@ class { 'redis::globals':
 
       context 'with ulimit' do
         let(:params) { { ulimit: 7777 } }
+        let(:facts) { super().merge(puppetversion: Puppet.version) }
```
ekohl: I thought rspec-puppet-facts did this for us:
https://github.com/mcanevet/rspec-puppet-facts/blob/e02496555b0a79c0c0434ed99557b7097c5771b4/lib/rspec-puppet-facts.rb#L237

fraenki: Good catch, fixed in most recent commit.

templates/service_templates/redis.service.erb
fraenki: Will do in a follow-up PR :) (see other comment)

spec/classes/redis_spec.rb
```diff
@@ -108,11 +117,20 @@ class { 'redis::globals':
...
+                with_incl("/etc/systemd/system/#{service_name}.service.d/limit.conf").
+                with_lens('Systemd.lns').
+                with_changes(['defnode nofile Service/LimitNOFILE ""', 'set $nofile/value "7777"']).
+                that_notifies('Class[systemd::systemctl::daemon_reload]')
+            else
+              is_expected.to contain_augeas('Systemd redis ulimit').
+                with_incl("/etc/systemd/system/#{service_name}.service.d/limit.conf").
+                with_lens('Systemd.lns').
+                with_changes(['defnode nofile Service/LimitNOFILE ""', 'set $nofile/value "7777"'])
+            end
```
ekohl: Alternative with less duplication
```puppet
            is_expected.to contain_augeas('Systemd redis ulimit').
              with_incl("/etc/systemd/system/#{service_name}.service.d/limit.conf").
              with_lens('Systemd.lns').
              with_changes(['defnode nofile Service/LimitNOFILE ""', 'set $nofile/value "7777"'])
            # Only necessary for Puppet < 6.1.0,
            # See https://github.com/puppetlabs/puppet/commit/f8d5c60ddb130c6429ff12736bfdb4ae669a9fd4
            if Puppet.version < '6.1'
              is_expected.to contain_augeas('Systemd redis ulimit').
                 that_notifies('Class[systemd::systemctl::daemon_reload]')
```

You can probably even add an else with `not_to`

PR #344, opened 2020-02-16
Add Debian 10 support

ekohl:
Closes GH-338

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -7,6 +7,7 @@
     - set: centos6-64m
     - set: centos7-64m
     - set: debian9-64m
+    - set: debian10-64m
 Gemfile:
   optional:
     ':test':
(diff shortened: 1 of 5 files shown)


ekohl:
https://github.com/voxpupuli/puppet-redis/pull/345 should fix the bolt issue. Will rebase on it once merged.

ghoneycutt:
@ekohl looks like it is failing on Debian 10 at https://github.com/voxpupuli/puppet-redis/blob/master/spec/spec_helper_acceptance.rb#L12

Seems the build is looking for mkdir in /usr/bin instead of /bin

ekohl:
https://github.com/voxpupuli/puppet-redis/pull/345 switches to using Bolt via the Puppet provided packages rather than Gems. Also updates from 0.5.1 to whatever is in the Puppet 5/Puppet 6 repos. That should fix it.

ekohl:
Currently fails because there's no dotdeb repository for Buster: https://github.com/voxpupuli/puppet-redis/pull/346

PR #345, opened 2020-02-16
Update bolt testing to use modern versions

ekohl:
This changes the bolt testing to use bolt from packages. Because the output is different, the tests are modified. It removes all helpers from spec_helper_acceptance and uses more native rspec syntax to simplify the testing.


ekohl:
It's finally green! Took a few tries :)

PR #346, opened 2020-02-17
Run acceptance tests with manage_repo => false

ekohl:
This makes sure we tests with the manifest defaults. It also allows testing on Debian 10 for which there is no dotdeb repository.


ekohl:
https://github.com/voxpupuli/puppet-redis/pull/347 submitted as a separate PR so the changelog actually captures the bugfix.

PR #350, opened 2020-04-01
Add support for RedHat/CentOS 8

yakatz:
Add support for RedHat/CentOS 8

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -6,6 +6,7 @@
     - set: ubuntu1804-64m
     - set: centos6-64m
     - set: centos7-64m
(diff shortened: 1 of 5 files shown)


ekohl:
Looks like this works, but the SCL tests need to be disabled since EL8 has no SCL.
https://github.com/voxpupuli/puppet-redis/blob/ef7c4e60f34af252e4a91207495adbbec756f020/spec/acceptance/suites/scl/redis5_spec.rb#L3

yakatz:
Tests all pass now.

ekohl:
Thanks!

PR #353, opened 2020-04-16
Correct config_dir for scl support

### LINKED ISSUE #341: Redhat SCL Config_Dir Issue

- Puppet: v2018.1.11 (Enterprise)
- Ruby:
- Distribution: RHEL 7
- Module version: 5.0.0

Configure the module utilizing Redhat Software Collections implementation for Redis and Sentinel installation and configuration. This is after updating the globals.pp $scl variable from 'undef' to "rh-redis5" so RHSCL is leveraged. 

## What are you seeing
After attempting to install and configure Sentinel on Redis instances, the pathing for the redis-sentinel.conf file does not appear to be getting referenced by Sentinel. Overwriting properties within the redis::sentinel, log_file as an example, will not impact Sentinel configurations. Puppet will update the conf file properly, however the running Sentinel instance will not see these changes.

Changes to the redis::sentinel class, without having to fully override the absolute path of the config_dir should apply to running instances of Sentinel.

## Output log
No logs that point this issue out, as on the surface it will appear to be a true configuration issue within your Redis/Sentinel setup.

We have resolved this issue internally by making the following code change. I believe one could also get around this by overriding the config_dir properties, however we felt that would lead to inconsistencies between the redis::sentinel and the redis classes and wanted a more cohesive/consistent solution. This puts all config files for both Redis and Redis Sentinel in the same location. 

Change description:
/redis/manifests/params.pp
Change line 60:
Old:  $config_dir                = "/etc/opt/rh/${scl}/redis"
New: $config_dir                = "/etc/opt/rh/${scl}"

markasammut:
Solves issue https://github.com/voxpupuli/puppet-redis/issues/341

Pull Request (PR) description

This Pull Request (PR) fixes the following issues

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -65,7 +65,7 @@ class redis::params inherits redis::globals {
 
       $scl = $redis::globals::scl
       if $scl {
-        $config_dir                = "/etc/opt/rh/${scl}/redis"
+        $config_dir                = "/etc/opt/rh/${scl}"
         $config_file               = "/etc/opt/rh/${scl}/redis.conf"
         $config_file_orig          = "/etc/opt/rh/${scl}/redis.conf.puppet"
         $package_name              = "${scl}-redis"


ekohl:
Could you come up with an acceptance test for this? Ideally one that fails before this patch and passes after. https://github.com/voxpupuli/puppet-redis/blob/master/spec/acceptance/suites/scl/redis5_spec.rb is the one we have now for Redis itself using SCL. https://github.com/voxpupuli/puppet-redis/blob/master/spec/acceptance/suites/default/redis_sentinel_one_node_spec.rb is the non-SCL Sentinel test.

markasammut:
Will try to, but I'm not that familiar with acceptance tests and spec.

ekohl:
If you're on Freenode, we are on #voxpupuli as well and I don't mind pointing you in the right direction. PR comments are also fine.

markasammut:
@ekohl Could this be merged please? It is a very simple change which would solve a number of issues. Thanks!

ekohl (approved):
I would have preferred to have a test for this, but sure.

PR #360, opened 2020-05-26
Add systemd module to metadata deps

tobias-urdin:
Adds the systemd module to the metadata
dependencies.


tobias-urdin:
Ah, didn't see it there, no worries then just thought it was a miss in documentation. I wasn't able to use it without that module which I was hoping for.

PR #362, opened 2020-07-25
modulesync 3.0.0 & puppet-lint updates

bastelfreak:
modulesync 3.0.0


ekohl (approved):
I added commits that I think should fix the tests. :+1: once it's green.

ekohl:
Manually removed Ubuntu 20.04 + Puppet 5 from testing since there are no packages for that.

PR #365, opened 2020-09-16
enable sentinel bind usage with multi ip

ragingdave:
There is currently an issue where if you use the allowed array of IP for the sentinel bind, this fails and the template becomes something completely broken as the resulting template shows:
```
bind ["10.0.0.155", "127.0.0.1"]
```

This PR fixes that by using the same technique in the base redis usage to force it into an array and flatten it to generate a correct bind line.

### DIFF
--- a/manifests/sentinel.pp
+++ b/manifests/sentinel.pp
@@ -149,7 +149,7 @@ class redis::sentinel (
     }
   }
 
-  $sentinel_bind_arr = [$sentinel_bind].flatten
+  $sentinel_bind_arr = delete_undef_values([$sentinel_bind]).flatten
 
   file { $config_file_orig:
     ensure  => file,
--- a/manifests/sentinel.pp
+++ b/manifests/sentinel.pp
@@ -149,7 +149,7 @@ class redis::sentinel (
     }
   }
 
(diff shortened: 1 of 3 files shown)


ekohl (commented):
The type also allows undef and this ends up with an empty `bind` statement and breaks the tests. Please have a look at that.

manifests/sentinel.pp
```diff
@@ -149,6 +149,8 @@
     }
   }
 
+  $sentinel_bind_arr = [$sentinel_bind].flatten
```
igalic, line 152: see delete_undef_values for @ekohl's comment, or it's documentation for how to do that with filter

igalic (changes requested):

bastelfreak:
@ragingdave thanks for the PR! Can you add a unit test for this change?

bastelfreak (approved):

manifests/sentinel.pp
```diff
@@ -149,6 +149,8 @@
     }
   }
 
+  $sentinel_bind_arr = delete_undef_values([$sentinel_bind]).flatten
```
igalic, line 152: it might make more sense to first `flatten()` and then `delete_undef_values()`

PR #272, opened 2018-08-28
Make protected-mode configurable for sentinel

eliskovets:

PR #367, opened 2020-09-22
release 6.2.0

bastelfreak:

PR #368, opened 2020-10-08
fix broken unit tests

bastelfreak:
chronological order:
* https://github.com/voxpupuli/puppet-redis/pull/272 was opened
* https://github.com/voxpupuli/puppet-redis/pull/365 was opened
* #365 was merged
* #272 was merged without a rebase
* tests are broken

### DIFF
--- a/manifests/sentinel.pp
+++ b/manifests/sentinel.pp
@@ -139,8 +139,22 @@ class redis::sentinel (
   Stdlib::Absolutepath $working_dir = $redis::params::sentinel_working_dir,
   Optional[Stdlib::Absolutepath] $notification_script = undef,
   Optional[Stdlib::Absolutepath] $client_reconfig_script = undef,
+  String[5] $minimum_version = $redis::params::minimum_version,
 ) inherits redis::params {
   require 'redis'
+
+  if $package_ensure =~ /^([0-9]+:)?[0-9]+\.[0-9]/ {
+    if ':' in $package_ensure {
+      $_redis_version_real = split($package_ensure, ':')
+      $redis_version_real = $_redis_version_real[1]
+    } else {
+      $redis_version_real = $package_ensure
+    }
+  } else {
+    $redis_version_real = pick(getvar('redis_server_version'), $minimum_version)
+  }
+
+  $supports_protected_mode = !$redis_version_real or versioncmp($redis_version_real, '3.2.0') >= 0
 
   if $facts['os']['family'] == 'Debian' {
     package { $package_name:
--- a/spec/classes/redis_sentinel_spec.rb
+++ b/spec/classes/redis_sentinel_spec.rb
@@ -136,6 +136,7 @@ port 26379
 dir /tmp/redis
 daemonize #{facts[:osfamily] == 'RedHat' ? 'no' : 'yes'}
 pidfile #{pidfile}
+protected-mode yes
 
 sentinel monitor cow 127.0.0.1 6379 2
 sentinel down-after-milliseconds cow 6000
--- a/templates/redis-sentinel.conf.erb
+++ b/templates/redis-sentinel.conf.erb
@@ -5,7 +5,9 @@ port <%= @sentinel_port %>
 dir <%= @working_dir %>
 daemonize <%= @daemonize ? 'yes' : 'no' %>
 pidfile <%= @pid_file %>
+<% if @supports_protected_mode -%>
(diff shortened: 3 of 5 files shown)


bastelfreak:
merge this, rebase and merge https://github.com/voxpupuli/puppet-redis/pull/367 afterwards

bastelfreak:
okay I'm not sure why the tests fail on Ubuntu 16 :(

spec/classes/redis_sentinel_spec.rb
```diff
@@ -136,6 +136,7 @@
 dir /tmp/redis
 daemonize #{facts[:osfamily] == 'RedHat' ? 'no' : 'yes'}
 pidfile #{pidfile}
+protected-mode yes
```
ekohl, line 139: You need to honor `$redis::sentinel::protected_mode`. The version on Ubuntu 16.04 is too old.

bastelfreak: mhm, we default to true in init.pp. you suggest it should be false for Ubuntu 16.04?

bastelfreak: not sure if this fix is what you had in mind, but it works now.

templates/redis-sentinel.conf.erb
```diff
@@ -5,7 +5,9 @@ port <%= @sentinel_port %>
 dir <%= @working_dir %>
 daemonize <%= @daemonize ? 'yes' : 'no' %>
 pidfile <%= @pid_file %>
+<% if @facts['os']['release']['major'] != '16.04' -%>
```
ekohl, line 8: This is not what I had in mind. This is much more like it:
https://github.com/voxpupuli/puppet-redis/blob/d44386133a7c524ba0f31d375cf91533f78db1cc/manifests/instance.pp#L402

manifests/sentinel.pp
```diff
@@ -153,8 +153,6 @@
     }
   }
 
-  $sentinel_bind_arr = delete_undef_values([$sentinel_bind].flatten)
```
ekohl, line 156: Was removing this intentional?

alexjfisher: I'm guessing not as the template is using this.

alexjfisher:
I'm going to knock up an alternative PR

manifests/sentinel.pp
```diff
@@ -139,9 +139,23 @@
   Stdlib::Absolutepath $working_dir = $redis::params::sentinel_working_dir,
   Optional[Stdlib::Absolutepath] $notification_script = undef,
   Optional[Stdlib::Absolutepath] $client_reconfig_script = undef,
+  String[5] $minimum_version = $redis::params::minimum_version,
 ) inherits redis::params {
   require 'redis'
 
+  if $package_ensure =~ /^([0-9]+:)?[0-9]+\.[0-9]/ {
+    if ':' in $package_ensure {
```
bastelfreak, line 147: I had to duplicate the whole block because $package_ensure for sentinel can be something differen than in init.pp/instance.pp

manifests/sentinel.pp
```diff
@@ -139,9 +139,23 @@
   Stdlib::Absolutepath $working_dir = $redis::params::sentinel_working_dir,
   Optional[Stdlib::Absolutepath] $notification_script = undef,
   Optional[Stdlib::Absolutepath] $client_reconfig_script = undef,
+  String[5] $minimum_version = $redis::params::minimum_version,
```
alexjfisher, line 142: Like in `instance.pp`, I think this is considered private and should just be in the class body instead.

basti-nis:
I've get the tests to work with following workaround in the template:

```
<% if @protected_mode -%>
protected-mode <%= @protected_mode ? 'yes' : 'no' %>
<% end -%>
```

i know, it's doubled if statement but i want to have it for later use, when the final solution exists.

the default value for protected mode is "no" if not set, so it's also fine for newer versions starting from 3.2.0.

let's give the option to the user, he should know how to and when to use this.

My workaround is implemented in  #356 (sorry for the massive commits :( )

basti-nis:
I think about it a little bit.

When protected mode is enabled by default, then you can write the config only if we want to disable it.
Then you can also check for OS Version or Redis version also.
What do you guys think about that?

ekohl:
In https://github.com/voxpupuli/puppet-redis/pull/374 I took an alternative approach.

PR #369, opened 2020-10-10
Make `sentinel::protected_mode` `Optional`

alexjfisher:
This might be a bit less disruptive than defaulting it to `true`, (and
means it won't break any sentinel 3.0 users if we've got any).

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -25,7 +25,6 @@ class redis::params inherits redis::globals {
       $sentinel_package_name     = 'redis-sentinel'
       $sentinel_log_file         = '/var/log/redis/redis-sentinel.log'
       $sentinel_working_dir      = '/var/lib/redis'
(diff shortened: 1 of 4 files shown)


ekohl:
I'd prefer to copy/share the approach for the main service:
https://github.com/voxpupuli/puppet-redis/blob/d44386133a7c524ba0f31d375cf91533f78db1cc/templates/redis.conf.erb#L23-L43 https://github.com/voxpupuli/puppet-redis/blob/d44386133a7c524ba0f31d375cf91533f78db1cc/manifests/instance.pp#L402

Distros default to protected-mode and it's a good idea to use it when available.

ekohl:
Looks like tests in current master are red - any plans to finish this?

alexjfisher:
This wasn’t the right fix.

PR #371, opened 2020-11-04
Drop dotdeb repo management support

ekohl:
The dotdeb project has been discontinued and never published packages for Debian 9 or 10. 645229f8aa507eb3213de0c4daa4cb8966859287 dropped Debian 8 support so this is essentially dead code that can only lead to problems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ class { '::redis':
 }
 ```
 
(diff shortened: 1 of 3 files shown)


kenyon (approved):

ekohl:
master is failing in the same way, merging.

ekohl:
Marking this as a bugfix since version 5.0.0 of the module dropped support for Debian 8. There never was a dotdeb repo for Debian 9 or newer so this was really not a feature that worked.

PR #374, opened 2020-11-12
Move package parameters to init.pp

ekohl:
This moves package_ensure and minimum_version parameters to init.pp.  They don't make sense on the instance because they're not actually honored. They are only used to determine whether protected mode is supported. Having that variable in the main redis class allows the sentinel to reuse the definition.


ekohl:
Except for Rubocop it passed. Updated now.

bastelfreak (approved):

alexjfisher (approved):

PR #376, opened 2020-11-18
add sentinel multi monitor

### LINKED ISSUE #223: Add Support for Multiple Sentinel Instances
Now that there is support for multiple Redis instances, there should also be the ability to add multiple Sentinel instances.

Requesting a feature request to support multiple Sentinel instances

basti-nis:
After multi instance was added we should let redis-sentinel watch over these instances

#### This Pull Request (PR) add following features and fixes the following issues
- multi instance support for redis-sentinel

This will close following Issues:
Close #375 
Close #223

### DIFF
--- a/manifests/sentinel.pp
+++ b/manifests/sentinel.pp
@@ -1,7 +1,10 @@
 # @summary Install redis-sentinel
 #
-# @param auth_pass
-#   The password to use to authenticate with the master and slaves.
+# @param sentinel_monitor
+#   Specify the sentinel monitor.
+#
+# @param monitor_defaults
+#   Override the monitor defaults
 #
 # @param config_file
 #   The location and name of the sentinel config file.
@@ -22,14 +25,6 @@
 # @param daemonize
 #   Have Redis sentinel run as a daemon.
 #
-# @param down_after
-#   Number of milliseconds the master (or any attached slave or sentinel)
-#   should be unreachable (as in, not acceptable reply to PING, continuously,
-#   for the specified period) in order to consider it in S_DOWN state.
-#
-# @param failover_timeout
-#   Specify the failover timeout in milliseconds.
-#
 # @param init_script
 #   Specifiy the init script that will be created for sentinel.
 #
@@ -38,16 +33,6 @@
 #
 # @param log_level
 #   Specify how much we should log.
-#
-# @param master_name
-#   Specify the name of the master redis server.
-#   The valid charset is A-z 0-9 and the three characters ".-_".
-#
-# @param redis_host
-#   Specify the bound host of the master redis server.
-#
-# @param redis_port
-#   Specify the port of the master redis server.
 #
 # @param protected_mode
 #   Whether protected mode is enabled or not. Only applicable when no bind is set.
@@ -58,16 +43,8 @@
 # @param package_ensure
 #   Do we ensure this package.
 #
-# @param parallel_sync
-#   How many slaves can be reconfigured at the same time to use a
-#   new master after a failover.
-#
 # @param pid_file
 #   If sentinel is daemonized it will write its pid at this location.
-#
-# @param quorum
-#   Number of sentinels that must agree that a master is down to
-#   signal sdown state.
 #
 # @param sentinel_bind
 #   Allow optional sentinel server ip binding.  Can help overcome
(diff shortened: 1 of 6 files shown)


basti-nis:
Hey guys,
this is finally ready for review!
Please have a look @ekohl or @bastelfreak !

Thank you!

bastelfreak:
A lot of changes going on here and I'm not a redis expert. @ekohl you maybe?

types/sentinelmonitor.pp
```diff
@@ -0,0 +1,12 @@
+type Redis::SentinelMonitor = Hash[String,Struct[{
```
ekohl, line 1: It's slightly confusing that the type is singular (monitor) but the content is multiple.
```suggestion
type Redis::SentinelMonitors = Hash[String,Struct[{
```

manifests/sentinel.pp
```diff
@@ -92,53 +69,57 @@
...
-  String[1] $package_name = $redis::params::sentinel_package_name,
-  String[1] $package_ensure = 'present',
-  Integer[0] $parallel_sync = 1,
-  Stdlib::Absolutepath $pid_file = $redis::params::sentinel_pid_file,
-  Integer[1] $quorum = 2,
+  String[1] $init_template                    = 'redis/redis-sentinel.init.erb',
+  Redis::LogLevel $log_level                  = 'notice',
+  Stdlib::Absolutepath $log_file              = $redis::params::sentinel_log_file,
+  Redis::SentinelMonitor $sentinel_monitor    = $redis::params::sentinel_default_monitor,
+  $monitor_defaults                           = $redis::params::sentinel_monitor_defaults,
```
ekohl, line 111: Please add a data type for this

ekohl (commented):
I'm wondering if it should be a defined type instead (`redis::sentinel::instance` or something). It should concat all monitored instances. The reason I think about this is that it's becoming harder to monitor.

basti-nis:
> 
> 
> I'm wondering if it should be a defined type instead (`redis::sentinel::instance` or something). It should concat all monitored instances. The reason I think about this is that it's becoming harder to monitor.

@ekohl 
Good point, will change it to that name!
I'll go into the 2 suggestions in this week, hope we can merge it soon :)

basti-nis:
> 
> 
> I'm wondering if it should be a defined type instead (`redis::sentinel::instance` or something). It should concat all monitored instances. The reason I think about this is that it's becoming harder to monitor.

@ekohl 
In a past PR you mentioned, that we should not add concat to our dependencies and further more that i should do a loop in the template. Or do i get you wrong?

redis::sentinel::instance is missleading, because it's is not for sentinel instances.
This PR makes it possible that your sentinal can monitor multiple redis instances.

types/sentinelmonitor.pp
basti-nis: applied

manifests/sentinel.pp
basti-nis: data type added

ekohl:
>  In a past PR you mentioned, that we should not add concat to our dependencies and further more that i should do a loop in the template. Or do i get you wrong?

If it's all in the same class, concat is not needed. However, if you use a defined type I think it will be needed.

> redis::sentinel::instance is missleading, because it is not for sentinel instances.
> This PR makes it possible that your sentinel can monitor multiple redis instances.

`instance` is just a name I suggested. It was more about a concept that you have something that repeats itself. A defined type can make it easier to see that.

> I think it's fine as it is.
> Not much has changed in the template, only a loop was added.
> Why do you think it's harder to monitor, i can't follow your thoughts.
> If you could explain it to me in more detail, i would adapt it.

I think passing a complex hash to a class is a painful API that's harder to understand than calling multiple defined types. In the end it'll result in the same content on disk so it's more about a usability from a developers point of view.

Does that make sense and help to elaborate my view?

basti-nis:
@ekohl 
> I think passing a complex hash to a class is a painful API that's harder to understand than calling multiple defined types. In the end it'll result in the same content on disk so it's more about a usability from a developers point of view.
> 
> Does that make sense and help to elaborate my view?

Ok, that makes sense. Thanks for your detailed answer!

> If it's all in the same class, concat is not needed. However, if you use a defined type I think it will be needed.
> 
> `instance` is just a name I suggested. It was more about a concept that you have something that repeats itself. A defined type can make it easier to see that.

You mean i should write an defined resource, that we can do something like this?:

```
redis::sentinel::monitor { 'my_redis_cluster':
  redis_host       => '1.2.3.4',
  redis_port       => 6881,
  down_after       => 5000,
  failover_timeout => 12000,
  quorum           => 2,
  parallel_sync    => 1,
  auth_pass        => 'secret',
}
```

Sorry, i think i missunderstood you in the past.

gizmo15:
Hi,

I'm facing the same need, is there any updates ?

Thanks

basti-nis:
Hi @gizmo15 ,

sorry, i didn't have time to write the code in the past months.
Actually my branch from my forked project works in our production environment without any problems almos a year now:
https://github.com/basti-nis/puppet-redis/tree/feature/sentinel_multi_monitor

If you need a quick solution, give it a try.

I will look into it, to complete the code like @ekohl suggested it, in the next couple of days.

If anybody else have an idea, feel free to contribute :)

gizmo15:
Hi,
Thanks for you answer and your work! we will check that!

TheMeier:
@basti-nis This is now untouched for a couple of years. Therefore I close it. If you want to continue working on please feel free to re-open it.
Thank you for your contributions

PR #377, opened 2020-11-24
Make augeasproviders_sysctl a soft dependency

ekohl:
augeasproviders_core was already redundant since augeasproviders_sysctl already depended on it and wasn't used directly.

Since augeasproviders_sysctl is only used on Red Hat and Debian to set ulimit when using non-systemd platforms (EL6, Debian with manual change from defaults). Those account for a very small number of the deployments so this makes it a soft dependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,8 @@ class { '::redis::sentinel':
 
 When managing the repo, it either needs [puppetlabs/apt](https://forge.puppet.com/puppetlabs/apt) or [puppet/epel](https://forge.puppet.com/puppet/epel).
 
+For administration of sysctl it depends on [herculesteam/augeasproviders_sysctl](https://forge.puppet.com/herculesteam/augeasproviders_sysctl).
+
 ## `redis::get()` function
 
 This function is used to get data from redis.
--- a/manifests/administration.pp
+++ b/manifests/administration.pp
@@ -3,6 +3,8 @@
 # For disabling Transparent Huge Pages (THP), use separate module such as:
 # https://forge.puppet.com/modules/alexharvey/disable_transparent_hugepage
 #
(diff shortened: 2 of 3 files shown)


ekohl:
I just noticed it also uses a lens to set the systemd override.

README.md
```diff
@@ -139,6 +139,8 @@ This module requires [camptocamp/systemd](https://forge.puppet.com/camptocamp/sy
 
 When managing the repo, it either needs [puppetlabs/apt](https://forge.puppet.com/puppetlabs/apt) or [puppet/epel](https://forge.puppet.com/puppet/epel).
 
+For ulimit support on non-systemd it depends on [herculesteam/augeasproviders_sysctl](https://forge.puppet.com/herculesteam/augeasproviders_sysctl).
```
ghoneycutt, line 142: This should let the user know which versions are supported. Suggest a reference to the `.fixtures.yml` file to find out the versions that are tested with this module. That way when those numbers get bumped in the future, the README does not also have to be updated.

ekohl:
I'm still not 100% about this, but now that we've dropped sysvinit, this makes more sense.

ekohl:
Replaced by https://github.com/voxpupuli/puppet-redis/pull/398

ekohl:
Ok, now I do think this is the correct way forward.

README.md
ekohl: That's kind of tricky. In fixtures we only test with master. I'm not sure what a good solution is. Having it in README is pretty much guaranteed to go out of date so I'm leaning to keeping it for the user. Given we only use 3 properties (name, ensure and value) I think it's safe to assume the API will continue to be the same.

bastelfreak (approved):

PR #378, opened 2020-11-25
modulesync 4.0.0

bastelfreak:
modulesync 4.0.0


ekohl (changes requested):
Previous beaker sets declared a master, our generated ones don't.

ekohl (approved):
Made a change, I think it should now pass (except on Ubuntu 20.04 + Puppet 5).

PR #379, opened 2020-12-10
Set the correct log dir on RH SCL

ekohl:
The path when using an SCL is non-standard. Using the correct one prevents permission issues.


bastelfreak (commented):
ubuntu 20 puppet 5 fails because of missing packages, centos 6 fails because it's eol. I'm fine with merging it

bastelfreak (approved):

PR #380, opened 2020-12-14
Make spec_helper.rb managed again

ekohl:
This allows getting updates via modulesync again. It currently depends on https://github.com/voxpupuli/voxpupuli-test/pull/19 which is why it's a draft. This is also to see if that PR actually works.


ekohl:
I think this is ready now.

PR #381, opened 2020-12-14
Drop EL6 support

ekohl:
CentOS 6 is EOL and we can't test RHEL 6. That means there's no guarantee it continues to work.


alexjfisher (changes requested):
Could we drop support and testing, but leave in the code for now?

That was the consensus at the last PMC meeting and @dhollinger has drafted a new policy document in https://github.com/voxpupuli/voxpupuli.github.io/pull/232

ekohl:
I don't like having dead code. In fact, I'm thinking of completely dropping sysvinit code since after this, everything listed in metadata.json defaults to using systemd. This allows using the instance support built into systemd (and deployed by default on Debian), making the integration more native.

ekohl:
In https://github.com/voxpupuli/puppet-redis/pull/382 I went to the next level. I'll keep both open for now, but I think that one has my preference.

kenyon (approved):

bastelfreak:
closing this in favour of https://github.com/voxpupuli/puppet-redis/pull/382

PR #382, opened 2020-12-15
Drop sysinit support

ekohl:
CentOS 6 is EOL and we can't test RHEL 6. That means there's no guarantee it continues to work. Also, all other distros mentioned in metadata.json run systemd. This allows the module to be cleaned up.

For the sentinel, every supported distro provides a systemd unit file as part of the package. On Debian, it also provides a sysvinit file.  Including this in the module is not needed. In most cases, the systemd unit file would have already shadowed the sysvinit file.

Alternative to https://github.com/voxpupuli/puppet-redis/pull/381.


ekohl:
The next step would be to convert instance support to use systemd's `redis@$instance` support. Debian 10 and newer Ubuntu ship a `redis@.service` file by default and if present, that should be used instead. Combined with systemd drop ins it should make the system a lot more integrated and the module smaller.

bastelfreak (approved):

alexjfisher:
I'm actually OK with this, as I think there's a clear argument in this module that 'Maintaining the code prevents the module from moving forward'. (As is the wording in https://github.com/voxpupuli/voxpupuli.github.io/pull/232)

manifests/instance.pp
```diff
@@ -316,77 +305,48 @@
...
-      }
-    } else {
-      file { "/etc/init.d/${service_name}":
-        ensure  => file,
-        mode    => '0755',
-        content => template("redis/service_templates/redis.${facts['os']['family']}.erb"),
-      }
+    # Only necessary for Puppet < 6.1.0,
+    # See https://github.com/puppetlabs/puppet/commit/f8d5c60ddb130c6429ff12736bfdb4ae669a9fd4
+    if versioncmp($facts['puppetversion'],'6.1.0') < 0 {
```
alexjfisher, line 318: +1 for leaving this in for now.  It could catch people out. When we remove it, we could replace it with a note in the README, either with an example people can use in their profiles, or just a quick link to eg. this or another comment.

PR #384, opened 2021-01-03
#372 Ensure THP is disabled after reboot

### LINKED ISSUE #372: After a reboot THP (Transparent Huge Pages) are enabled again

current

```
contain '::redis::administration'

Class['::redis::administration']
-> Class['::redis']
```

## What are you seeing

On a first run THP are being disabled before Redis service is started, so it affects Redis.

However if you reboot the server then Redis will start with THP enabled as it's not disabled in a persistent way.

THP to be disabled in a way that will survive the server restart.

dom-nie:
Change the way THP is managed, by adding a way to disable it without running puppet in case of a reboot.

Fixes #372

### DIFF
--- a/files/service_files/disable_thp.service
+++ b/files/service_files/disable_thp.service
@@ -1,0 +1,9 @@
+[Unit]
+Description=Disables THP on boot
+
+[Service]
+Type=oneshot
+ExecStart=/bin/hugeadm --thp-never
+
+[Install]
+WantedBy=default.target
--- a/manifests/administration.pp
+++ b/manifests/administration.pp
@@ -36,21 +36,16 @@ class redis::administration (
     package { $hugeadm_package:
       ensure => 'present',
     }
-
-    exec { 'systemd run_once disable_thp':
-      command     => '/usr/bin/systemctl start disable_thp.service',
-      refreshonly => true,
-      subscribe   => Systemd::Unit_file['disable_thp.service'],
-    }
   }
 
-  $ensure_thp_service = $disable_thp ? { true => 'present', false => 'absent' }
+  $ensure_thp = $disable_thp ? { true => 'present', false => 'absent' }
 
-  systemd::unit_file { 'disable_thp.service':
-    ensure  => $ensure_thp_service,
-    active  => false,
-    enable  => $disable_thp,
-    content => file("${module_name}/service_files/disable_thp.service"),
+  systemd::timer { 'disable_thp.timer':
+    ensure          => $ensure_thp,
+    timer_content   => file("${module_name}/service_files/disable_thp.timer"),
+    service_content => file("${module_name}/service_files/disable_thp.service"),
+    active          => false,
+    enable          => $disable_thp,
   }
 
   if $somaxconn > 0 {
--- a/manifests/administration.pp
+++ b/manifests/administration.pp
@@ -23,50 +23,34 @@ class redis::administration (
   Boolean $enable_overcommit_memory = true,
   Boolean $disable_thp              = true,
   Integer[0] $somaxconn             = 65535,
-) {
+  String[1] $hugeadm_package        = $redis::params::hugeadm_package,
+) inherits redis::params {
   if $enable_overcommit_memory {
     sysctl { 'vm.overcommit_memory':
       ensure => 'present',
       value  => '1',
     }
   }
-  file { '/etc/systemd/system/disable_thp.service':
-    ensure  => file,
-    mode    => '0644',
-    owner   => 'root',
-    group   => 'root',
-    content => file('redis/service_files/disable_thp.service'),
+
+  if $disable_thp {
+    package { $hugeadm_package:
+      ensure => 'present',
+    }
+
+    exec { 'systemd run_once disable_thp':
+      command     => '/usr/bin/systemctl start disable_thp.service',
+      refreshonly => true,
+      subscribe   => Systemd::Unit_file['disable_thp.service'],
+    }
   }
 
-  # Only necessary for Puppet < 6.1.0,
-  # See https://github.com/puppetlabs/puppet/commit/f8d5c60ddb130c6429ff12736bfdb4ae669a9fd4
-  if versioncmp($facts['puppetversion'],'6.1.0') < 0 {
-    include systemd::systemctl::daemon_reload
-    File['/etc/systemd/system/disable_thp.service'] ~> Class['systemd::systemctl::daemon_reload']
-  }
+  $ensure_thp_service = $disable_thp ? { true => 'present', false => 'absent' }
 
-  $hugeadm_package = $facts['os']['family'] ? {
-    'RedHat' => 'libhugetlbfs-utils',
-    'Debian' => 'libhugetlbfs',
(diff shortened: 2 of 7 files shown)


bastelfreak:
Hi,
can you please rebase against our master branch? Since https://github.com/voxpupuli/puppet-redis/pull/382 got merged we the broken centos 6 tests disappear.

dom-nie:
The branch is rebased to the current master, and the failing test isn't able to download puppet5 deb for Ubuntu20.04.

manifests/administration.pp
```diff
@@ -30,14 +30,43 @@
       value  => '1',
     }
   }
+  file { '/etc/systemd/system/disable_thp.service':
+    ensure  => file,
+    mode    => '0644',
+    owner   => 'root',
+    group   => 'root',
+    content => file('redis/service_files/disable_thp.service'),
+  }
```
ekohl, line 39: We already depend the systemd module. Perhaps it should use `systemd::unit_file` instead?

manifests/administration.pp
```diff
@@ -30,14 +30,43 @@
...
+    ensure => 'present',
+  }
+
+  service { 'disable_thp':
+    ensure    => false,
+    enable    => $disable_thp,
+    subscribe => File['/etc/systemd/system/disable_thp.service'],
+  }
+
+  exec { 'systemd run_once disable_thp':
```
towo, line 65: AFAICT this exec will *always* run, even if `$disable_thp` is false. That's probably not supposed to happen?

In general, it seems a bit hacky to make a one-off subscribe to a file change. You might want to consider writing a quick fact that checks the status and ensures thp is disabled.

ekohl (commented):
Perhaps disable_thp could be split off into its own class? Especially if it installs a package that feels more natural.

manifests/administration.pp
```diff
@@ -30,14 +30,43 @@
...
+    'Archlinux' => 'libhugetlbfs',
+    default  => 'hugepages',
+  }
+
+  package { $hugeadm_package:
+    ensure => 'present',
+  }
+
+  service { 'disable_thp':
+    ensure    => false,
```
towo, line 60: Personal preference and very susceptible to being overruled by others, but 'false' is confusing and should be replaced with 'stopped'. It's the same in the provider source, so no change in function.

manifests/administration.pp
towo: Or, as I just checked up on docs, add a timer unit with OnBootSec=0 as a trigger, that *should* fire the moment you start the timer.

towo (changes requested):

manifests/administration.pp
```diff
@@ -30,14 +30,43 @@
...
-      unless  => 'cat /sys/kernel/mm/transparent_hugepage/enabled | grep "\[never\]"',
-    }
+  # Only necessary for Puppet < 6.1.0,
+  # See https://github.com/puppetlabs/puppet/commit/f8d5c60ddb130c6429ff12736bfdb4ae669a9fd4
+  if versioncmp($facts['puppetversion'],'6.1.0') < 0 {
+    include systemd::systemctl::daemon_reload
+    File['/etc/systemd/system/disable_thp.service'] ~> Class['systemd::systemctl::daemon_reload']
+  }
+
+  $hugeadm_package = $facts['os']['family'] ? {
```
bastelfreak, line 48: can you move the logic into the existing one in params.pp and make $hugeadm_package a parameter?

manifests/administration.pp
```diff
@@ -30,14 +30,43 @@
       value  => '1',
     }
   }
+  file { '/etc/systemd/system/disable_thp.service':
```
bastelfreak, line 33: The $disable_thp condition got lost and hugepages are now disabled by default. Please add it back

manifests/administration.pp
dom-nie: Done

manifests/administration.pp
dom-nie: Done

manifests/administration.pp
dom-nie: Done

manifests/administration.pp
dom-nie: Sure, I can do this

manifests/administration.pp
dom-nie: @towo  It was indeed hacked to ensure THP will be disabled, before reboot for initial setup of host.
I have introduced timer as you have suggested.

dom-nie:
@ekohl  Haven't seen this comment before, but I can do this since it seems valid. Do you think it should be a fully separate class from Administration, or simply included in it?

ekohl:
> Do you think it should be a fully separate class from Administration, or simply included in it?

I think it should be a class that can function standalone and, if desired, is then included in administration (ensuring the right parameters are set). Does that make sense?

files/service_files/disable_thp.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
```
bastelfreak, line 1: can you add a comment at the top like `# THIS FILE IS MANAGED BY PUPPET`? Same for the .timer file

files/service_files/disable_thp.service
```diff
@@ -0,0 +1,9 @@
...
+Description=Disables THP on boot
+
+[Service]
+Type=oneshot
+ExecStart=/bin/hugeadm --thp-never
```
bastelfreak, line 6: Iv'e the feeling that this path isn't the same on all operating systems, but I really don't know.

manifests/administration.pp
```diff
@@ -32,14 +33,21 @@
...
-      onlyif  => 'test -f /sys/kernel/mm/transparent_hugepage/enabled',
-      unless  => 'cat /sys/kernel/mm/transparent_hugepage/enabled | grep "\[never\]"',
+    package { $hugeadm_package:
+      ensure => 'present',
     }
   }
 
+  $ensure_thp = $disable_thp ? { true => 'present', false => 'absent' }
+
+  systemd::timer { 'disable_thp.timer':
```
bastelfreak, line 43: the systemd::timer resource also needs to go into the `if $disable_thp {}` block

files/service_files/disable_thp.service
towo: Correct, it's [`/usr/bin/hugeadm`](https://packages.debian.org/buster/amd64/libhugetlbfs-bin/filelist) on Debian, and it's not Arch where `/bin` is symlinked.

files/service_files/disable_thp.service
towo: Correct, it's [`/usr/bin/hugeadm`](https://packages.debian.org/buster/amd64/libhugetlbfs-bin/filelist) on Debian, and it's not Arch where `/bin` is symlinked.

manifests/administration.pp
towo: Alongside the `$ensure_thp` definition ofc

towo (approved):

towo (approved):
Pending the changes requested by @bastelfreak: LGTM.

PR #385, opened 2021-01-06
Remove support for THP in Redis module - recommend forge module

### LINKED ISSUE #372: After a reboot THP (Transparent Huge Pages) are enabled again

dom-nie:
After discussing #384 with @ekohl  I have reached the conclusion that instead of further complicating this module, it would be best to simplify it and recommend the forge module for disabling THP.
I greatly appreciate all the feedback.

Closes #372

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -9,7 +9,7 @@
 #### Public Classes
 
 * [`redis`](#redis): This class installs redis
-* [`redis::administration`](#redisadministration): Allows various adminstrative settings for Redis As documented in the FAQ and https://redis.io/topics/admin
+* [`redis::administration`](#redisadministration): Allows various administrative settings for Redis
 * [`redis::globals`](#redisglobals): Set a global config for Redis
 * [`redis::sentinel`](#redissentinel): Install redis-sentinel
 
@@ -29,7 +29,7 @@ repositories.
 
 ### Functions
 
-* [`redis::get`](#redisget): ```
+* [`redis::get`](#redisget): Returns the value of the key being looked up or `undef` if the key does not exist.
 
 ### Data types
 
@@ -886,8 +886,9 @@ Default value: `$redis::params::minimum_version`
 
 ### `redis::administration`
 
-Allows various adminstrative settings for Redis
-As documented in the FAQ and https://redis.io/topics/admin
+As documented in the FAQ and https://redis.io/topics/admin.
+For disabling Transparent Huge Pages (THP), use separate module such as:
+https://forge.puppet.com/modules/alexharvey/disable_transparent_hugepage
 
 * **See also**
   * https://redis.io/topics/admin
@@ -904,7 +905,7 @@ include redis::administration
 
 ```puppet
 class {'redis::administration':
-  disable_thp => false,
+  enable_overcommit_memory => false,
 }
 ```
 
@@ -917,14 +918,6 @@ The following parameters are available in the `redis::administration` class.
 Data type: `Boolean`
 
 Enable the overcommit memory setting
-
-Default value: ``true``
-
-##### `disable_thp`
-
-Data type: `Boolean`
-
-Disable Transparent Huge Pages
 
 Default value: ``true``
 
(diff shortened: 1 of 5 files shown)


REFERENCE.md
```diff
@@ -886,25 +886,28 @@ Default value: `$redis::params::minimum_version`
 
 ### `redis::administration`
 
-Allows various adminstrative settings for Redis
+Allows various administrative settings for Redis
 As documented in the FAQ and https://redis.io/topics/admin
 
+For disabling Transparent Huge Pages please use forge module:
```
gcoxmoz, line 892: nitpick/suggestion.  Your change eliminates `disable_thp`, and also all `thp` strings.  How about you make this line "For disabling Transparent Huge Pages (THP), please" etc etc.

No pressure, feel free to ignore.

kenyon (approved):

REFERENCE.md
dom-nie: I actually think you are totally right, so I have added THP to make sure it's easier to search for.

REFERENCE.md
```diff
@@ -886,25 +886,28 @@ Default value: `$redis::params::minimum_version`
 
 ### `redis::administration`
 
-Allows various adminstrative settings for Redis
+Allows various administrative settings for Redis
 As documented in the FAQ and https://redis.io/topics/admin
 
+For disabling Transparent Huge Pages (THP), please use forge module:
+https://github.com/alexharv074/puppet-disable_transparent_hugepage
```
alexjfisher, line 893: Maybe link to the forge module instead??
https://forge.puppet.com/modules/alexharvey/disable_transparent_hugepage

ekohl: I personally haven't used this module. Do want to endorse it in this way or tone it down to "use separate module such as ..."?

REFERENCE.md
```diff
@@ -886,25 +886,28 @@ Default value: `$redis::params::minimum_version`
 
 ### `redis::administration`
 
-Allows various adminstrative settings for Redis
+Allows various administrative settings for Redis
```
ekohl, line 889: You should fix the typo in the file itself and then regenerate `REFERENCE.md` using `bundle exec rake strings:generate:reference`.

ekohl (commented):
Overall :+1: for dropping this functionality.

REFERENCE.md
dom-nie: Okay, moved the comment regarding THP to the class itself. Generated reference file but is this many changes expected @ekohl ?

REFERENCE.md
dom-nie: Oh, I used both of your recommendations here.

REFERENCE.md
```diff
@@ -1,19 +1,18 @@
...
-
 <!-- DO NOT EDIT: This document was generated by Puppet Strings -->
 
 ## Table of Contents
 
-### Classes
+**Classes**
 
-#### Public Classes
+_Public Classes_
```
ekohl, line 8: Could it be that you have an old version of strings installed and that it's using an old template?

manifests/administration.pp
```diff
@@ -1,18 +1,18 @@
-# Allows various adminstrative settings for Redis
-# As documented in the FAQ and https://redis.io/topics/admin
+# Allows various administrative settings for Redis
```
ekohl, line 1: Perhaps this should be an explicit summary to keep it short.
```suggestion
# @summary Allows various administrative settings for Redis
```

REFERENCE.md
dom-nie: You are right, I have updated this pull request with the latest version.

ekohl (approved):
Looks good to me. Should we squash this in a single commit? Can be done on merge.

dom-nie:
@ekohl  Thanks for all the review. I changed it to a single commit.

PR #387, opened 2021-01-27
Allow relative paths for $log_file

ekohl:
0c851ce3cd8d44256a3577dd81eaa88614095ac4 changed $log_dir for SCL deployments. However, $log_file was still absolute. That caused it to log to a non-existing directory.

This patch makes $log_file accept a relative path. If a relative path is specified, $log_dir will be prepended. Absolute paths are still used directly which keeps API compatibility.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -82,7 +82,8 @@
 # @param log_dir_mode
 #   Adjust mode for directory containing log files.
 # @param log_file
-#   Specify file where to write log entries.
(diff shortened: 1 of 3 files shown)


ehelms (approved):
Fixed issue caused by the previous commit as seen in https://github.com/theforeman/puppet-pulpcore/pull/169

bastelfreak (approved):

ekohl:
Hopefully now with green tests.

bastelfreak (approved):

PR #388, opened 2021-01-31
Adding support for Redis6 config and new features

dom-nie:
This commit introduces the latest config for Redis6 and Sentinel, with new parameters. Additionally, the MemoryPolicy type is introduced with some small adjustments in the rest of the code.

As discussed previously on voxpupuli IRC this is a base commit, and depending on the review I will make changes accordingly.


templates/redis6-sentinel.conf.erb
```diff
@@ -0,0 +1,334 @@
+# *** IMPORTANT ***
```
bastelfreak, line 1: can you please convert the template to epp? We would like to get rid of erb on the long run and to achieve this, all new templates have to be epp.

REFERENCE.md
```diff
@@ -42,849 +43,1033 @@ repositories.
 
 ## Classes
 
-### `redis`
+### <a name="redis"></a>`redis`
```
ekohl, line 46: Could you regenerate `REFERENCE.md` prior to the changes in a separate commit to make reviewing easier?

spec/classes/redis_spec.rb
```diff
@@ -679,7 +679,7 @@ class { 'redis::globals':
 
         it do
           if facts[:operatingsystem] == 'Ubuntu' && facts[:operatingsystemmajrelease] == '16.04'
-            is_expected.not_to contain_file(config_file_orig).with_content(%r{protected-mode})
+            is_expected.not_to contain_file(config_file_orig).with_content(%r{^protected-mode$})
```
ekohl, line 682: The regex isn't really needed. In fact, I'd say it's better not to be strict. It's expecting that `protected-mode` isn't present at all, regardless of where in the file it shows up.

spec/classes/redis_spec.rb
```diff
@@ -43,8 +43,8 @@
 
         it do
           is_expected.to contain_service(service_name).with(
-            'ensure'     => 'running',
-            'enable'     => 'true',
+            'ensure' => 'running',
+            'enable' => 'true',
```
ekohl, line 47: Generally it makes it easier to review if you separate out the visual changes to a separate commit.

templates/redis6-sentinel.conf.erb
dom-nie: @bastelfreak Done

REFERENCE.md
dom-nie: @ekohl I did the split to two commits

spec/classes/redis_spec.rb
dom-nie: I would say it's better to be strict.
If we only set variables, without touching documentation inside the file, so people can use them for reference then 
`# protected-mode example` would be shown somewhere which is caught by the previous regexp and the test fails

spec/classes/redis_spec.rb
dom-nie: Removed from the scope of this changes

ekohl:
This now has a merge conflict.

PR #391, opened 2021-02-22
Manage ulimit with systemd::dropin_file and allow it to be unmanaged

### LINKED ISSUE #390: Make ulimit optional or use puppet-systemd for dropin
This module currently contains `redis::ulimit` on every run (https://github.com/voxpupuli/puppet-redis/blob/master/manifests/config.pp#L41-L43) which can't be disabled due to the parameter constraints here: https://github.com/voxpupuli/puppet-redis/blob/master/manifests/init.pp#L318

This makes it impossible to add an additional `systemd::dropin` using https://github.com/camptocamp/puppet-systemd as the dropin directory get's enforced by `redis::ulimit` and also `systemd::dropin_file` (= duplicate declaration).

Two possible solutions could be:
- Fix the parameter definition to allow `undef` and therefore skip the containment of `redis::ulimit` -> A user would have to explicitly define the ulimit
- Switch to puppet-systemd for managing the dropin directory and the dropin itself (instead of `file` and `augeas`).

- Puppet: any
- Ruby: any
- Distribution: any
- Module version: any

```
class { '::redis':
    bind => ['127.0.0.1', $::facts['default_ipaddress']],
  }

  systemd::dropin_file { 'redis-dropin2.conf':
    unit     => 'redis.service',
    content  => '[Unit]\nDescription=Testdescription',
  }
```
## What are you seeing
```sh
Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Function Call, Duplicate declaration: File[/etc/systemd/system/redis.service.d] is already declared
```
Two dropin files managed by puppet

laugmanuel:
This PR changes two things:
- Adds the new parameter `ulimit_managed` to explicitly enable/disable ulimit
  - in systemd unit für instances and
  - systemd dropin for default setup
- Use [systemd:: service_limits](https://github.com/camptocamp/puppet-systemd/blob/master/manifests/service_limits.pp) to manage the ulimits for the default setup.
  Note: This adds a new dependency to [camptocamp/puppet-systemd](https://github.com/camptocamp/puppet-systemd).
- Use [systemd::unit_file](https://github.com/camptocamp/puppet-systemd/blob/master/manifests/unit_file.pp) for each instance service

Fixes #390

Note: I can split this PR into two/three PRs if desired (or squash the commits together).

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,8 +135,6 @@ class { '::redis::sentinel':
 
 ### Soft dependency
 
-This module requires [camptocamp/systemd](https://forge.puppet.com/camptocamp/systemd) on Puppet versions older than 6.1.0.
-
 When managing the repo, it either needs [puppetlabs/apt](https://forge.puppet.com/puppetlabs/apt) or [puppet/epel](https://forge.puppet.com/puppet/epel).
 
 ## `redis::get()` function
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -195,6 +195,8 @@
 #   Close the connection after a client is idle for N seconds (0 to disable).
 # @param ulimit
 #   Limit the use of system-wide resources.
+# @param ulimit_managed
+#   Defines wheter the Limit is managed by this module or not.
 # @param unixsocket
 #   Define unix socket path
 # @param unixsocketperm
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -316,6 +316,7 @@ class redis (
   Variant[Stdlib::Absolutepath, Enum['']] $unixsocket            = '/var/run/redis/redis.sock',
   Variant[Stdlib::Filemode, Enum['']] $unixsocketperm            = '0755',
   Integer[0] $ulimit                                             = 65536,
+  Boolean $ulimit_managed                                        = true,
   Stdlib::Absolutepath $workdir                                  = $redis::params::workdir,
   Stdlib::Filemode $workdir_mode                                 = '0750',
   Integer[0] $zset_max_ziplist_entries                           = 128,
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -196,7 +196,8 @@
 # @param ulimit
 #   Limit the use of system-wide resources.
 # @param ulimit_managed
-#   Defines wheter the Limit is managed by this module or not.
+#   Defines wheter the max number of open files for the
+#   systemd service unit is explicitly managed.
 # @param unixsocket
 #   Define unix socket path
 # @param unixsocketperm
--- a/manifests/instance.pp
+++ b/manifests/instance.pp
@@ -164,7 +164,7 @@
 # @param ulimit
 #   Limit the use of system-wide resources.
 # @param ulimit_managed
-#   Defines wheter the max number of open files for the
+#   Defines whether the max number of open files for the
 #   systemd service unit is explicitly managed.
 # @param unixsocket
 #   Define unix socket path
--- a/manifests/instance.pp
+++ b/manifests/instance.pp
@@ -163,6 +163,9 @@
 #   Close the connection after a client is idle for N seconds (0 to disable).
 # @param ulimit
 #   Limit the use of system-wide resources.
+# @param ulimit_managed
+#   Defines wheter the max number of open files for the
+#   systemd service unit is explicitly managed.
 # @param unixsocket
 #   Define unix socket path
 # @param unixsocketperm
@@ -310,8 +313,8 @@ define redis::instance (
   }
 
   if $manage_service_file {
-    file { "/etc/systemd/system/${service_name}.service":
-      ensure  => file,
+    systemd::unit_file { "${service_name}.service":
+      ensure  => 'present',
       owner   => 'root',
       group   => 'root',
       mode    => '0644',
--- a/manifests/instance.pp
+++ b/manifests/instance.pp
@@ -263,6 +263,7 @@ define redis::instance (
   Integer[0] $timeout                                            = $redis::timeout,
   Variant[Stdlib::Filemode , Enum['']] $unixsocketperm           = $redis::unixsocketperm,
   Integer[0] $ulimit                                             = $redis::ulimit,
+  Boolean $ulimit_managed                                        = $redis::ulimit_managed,
   Stdlib::Filemode $workdir_mode                                 = $redis::workdir_mode,
   Integer[0] $zset_max_ziplist_entries                           = $redis::zset_max_ziplist_entries,
   Integer[0] $zset_max_ziplist_value                             = $redis::zset_max_ziplist_value,
--- a/manifests/ulimit.pp
+++ b/manifests/ulimit.pp
@@ -31,7 +31,7 @@ class redis::ulimit {
     limits          => {
       'LimitNOFILE' => $redis::ulimit,
     },
-    restart_service =>  false,
+    restart_service => false,
   }
 
   file { "/etc/systemd/system/${redis::service_name}.service.d/limit.conf":
--- a/manifests/ulimit.pp
+++ b/manifests/ulimit.pp
@@ -27,31 +27,14 @@ class redis::ulimit {
     }
   }
 
-  file { "/etc/systemd/system/${redis::service_name}.service.d/":
-    ensure                  => 'directory',
-    owner                   => 'root',
-    group                   => 'root',
-    selinux_ignore_defaults => true,
+  systemd::service_limits { "${redis::service_name}.service":
+    limits          => {
+      'LimitNOFILE' => $redis::ulimit,
+    },
+    restart_service =>  false,
   }
 
   file { "/etc/systemd/system/${redis::service_name}.service.d/limit.conf":
-    ensure => file,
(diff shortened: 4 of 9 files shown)


manifests/init.pp
```diff
@@ -316,6 +316,7 @@
   Variant[Stdlib::Absolutepath, Enum['']] $unixsocket            = '/var/run/redis/redis.sock',
   Variant[Stdlib::Filemode, Enum['']] $unixsocketperm            = '0755',
   Integer[0] $ulimit                                             = 65536,
+  Boolean $ulimit_managed                                        = true,
```
ekohl, line 319: Could you add documentation for this?

manifests/ulimit.pp
```diff
@@ -27,31 +27,8 @@
...
-    ],
-  }
-  # Only necessary for Puppet < 6.1.0,
-  # See https://github.com/puppetlabs/puppet/commit/f8d5c60ddb130c6429ff12736bfdb4ae669a9fd4
-  if versioncmp($facts['puppetversion'],'6.1.0') < 0 {
-    include systemd::systemctl::daemon_reload
-    Augeas['Systemd redis ulimit'] ~> Class['systemd::systemctl::daemon_reload']
+  systemd::dropin_file { 'limit.conf':
+    unit    => "${redis::service_name}.service",
+    content => "# File managed by Puppet\n[Service]\nLimitNOFILE=${redis::ulimit}",
```
ekohl, line 32: Files with end of lines are generally better
```suggestion
    content => "# File managed by Puppet\n[Service]\nLimitNOFILE=${redis::ulimit}\n",
```

manifests/ulimit.pp
```diff
@@ -27,31 +27,8 @@
...
-      "defnode nofile Service/LimitNOFILE \"\"",
-      "set \$nofile/value \"${redis::ulimit}\"",
-    ],
-  }
-  # Only necessary for Puppet < 6.1.0,
-  # See https://github.com/puppetlabs/puppet/commit/f8d5c60ddb130c6429ff12736bfdb4ae669a9fd4
-  if versioncmp($facts['puppetversion'],'6.1.0') < 0 {
-    include systemd::systemctl::daemon_reload
-    Augeas['Systemd redis ulimit'] ~> Class['systemd::systemctl::daemon_reload']
+  systemd::dropin_file { 'limit.conf':
```
ekohl, line 30: Does it make sense to use [systemd::service_limits](https://github.com/camptocamp/puppet-systemd/blob/master/manifests/service_limits.pp) instead?

laugmanuel:
Test failures seem to be unrelated...

ekohl (changes requested):
Perhaps this should also be modified to use `systemd::unit_file`:
https://github.com/voxpupuli/puppet-redis/blob/551e58b50dd6101b9ad563b81de07aafaa2dfdb1/manifests/instance.pp#L312-L318

https://github.com/voxpupuli/puppet-redis#soft-dependency also needs an update if systemd becomes a hard dependency.

manifests/instance.pp
```diff
@@ -263,6 +263,7 @@
   Integer[0] $timeout                                            = $redis::timeout,
   Variant[Stdlib::Filemode , Enum['']] $unixsocketperm           = $redis::unixsocketperm,
   Integer[0] $ulimit                                             = $redis::ulimit,
+  Boolean $ulimit_managed                                        = $redis::ulimit_managed,
```
ekohl, line 266: Could you also document this?

manifests/init.pp
```diff
@@ -195,6 +195,8 @@
 #   Close the connection after a client is idle for N seconds (0 to disable).
 # @param ulimit
 #   Limit the use of system-wide resources.
+# @param ulimit_managed
+#   Defines wheter the Limit is managed by this module or not.
```
ekohl, line 199: Maybe I'm to blame for this, but perhaps it's a good chance to clear up the text? As a user this doesn't really tell me what it does.

manifests/ulimit.pp
```diff
@@ -27,31 +27,14 @@
...
-      "set \$nofile/value \"${redis::ulimit}\"",
-    ],
-  }
-  # Only necessary for Puppet < 6.1.0,
-  # See https://github.com/puppetlabs/puppet/commit/f8d5c60ddb130c6429ff12736bfdb4ae669a9fd4
-  if versioncmp($facts['puppetversion'],'6.1.0') < 0 {
-    include systemd::systemctl::daemon_reload
-    Augeas['Systemd redis ulimit'] ~> Class['systemd::systemctl::daemon_reload']
+    ensure => absent,
   }
```
ekohl, line 39: There is no daemon reload code for Puppet < 6.1.0 in `systemd::service_limits`. Interestingly, in combination with `restart_service` that appears to be a bug.

README.md
```diff
@@ -133,9 +133,11 @@ class { '::redis::sentinel':
 }
 ```
 
-### Soft dependency
+### Hard dependency
```
ekohl, line 136: Generally we don't document hard dependencies. They're already in metadata.json anyway.

manifests/ulimit.pp
laugmanuel: The daemon reload is contained in `systemd::dropin_file` which is invoked indirectly by `systemd::service_limits`. As this PR would use the new way of defining the limits regardless of Puppet version, the reload should be handled there.

manifests/instance.pp
```diff
@@ -309,8 +313,8 @@
   }
 
   if $manage_service_file {
-    file { "/etc/systemd/system/${service_name}.service":
-      ensure  => file,
+    systemd::unit_file { "${service_name}.service":
```
ekohl, line 316: If you pass enable and active as well, the unit file can manage the service and also properly apply `daemon_reload`. This is particularly important now that `camptocamp/systemd` has dropped the `daemon_reload` class in master.

manifests/ulimit.pp
```diff
@@ -27,31 +27,14 @@
...
-      "defnode nofile Service/LimitNOFILE \"\"",
-      "set \$nofile/value \"${redis::ulimit}\"",
-    ],
-  }
-  # Only necessary for Puppet < 6.1.0,
-  # See https://github.com/puppetlabs/puppet/commit/f8d5c60ddb130c6429ff12736bfdb4ae669a9fd4
-  if versioncmp($facts['puppetversion'],'6.1.0') < 0 {
-    include systemd::systemctl::daemon_reload
-    Augeas['Systemd redis ulimit'] ~> Class['systemd::systemctl::daemon_reload']
+    ensure => absent,
```
ekohl, line 38: Thinking more about this, it's a bit dubious how this should be handled. Previously it was an augeas lens which means it did manage one entry but users were free to manage other entries via other means.

laugmanuel: The idea was that there is no duplicate of old and new settings. But you are right; if users added their custom limits, they should not get removed.
We could switch the augeas lens to exclude the the NoFile line or just stop managing the file at all (which will result in duplicates of NOFile)

laugmanuel: Or we just stick to using the current approach and I'll drop switching to puppet-systemd from this PR.

manifests/instance.pp
```diff
@@ -163,6 +163,9 @@
 #   Close the connection after a client is idle for N seconds (0 to disable).
 # @param ulimit
 #   Limit the use of system-wide resources.
+# @param ulimit_managed
+#   Defines wheter the max number of open files for the
```
ekohl, line 167:
```suggestion
#   Defines whether the max number of open files for the
```

manifests/ulimit.pp
ekohl: I don't mind doing a major version bump for it and making it clear that this is the behavior.

manifests/instance.pp
ekohl: See https://github.com/voxpupuli/puppet-redis/commit/58aa20ccded225c5acab3cd2ad24f6e595f6632a for what I think is roughly needed.

manifests/ulimit.pp
```diff
@@ -27,31 +27,14 @@
     }
   }
 
-  file { "/etc/systemd/system/${redis::service_name}.service.d/":
-    ensure                  => 'directory',
-    owner                   => 'root',
-    group                   => 'root',
-    selinux_ignore_defaults => true,
+  systemd::service_limits { "${redis::service_name}.service":
```
ekohl, line 30: What do you think about moving it into instance.pp? https://github.com/voxpupuli/puppet-redis/commit/2c2a3223aa1338985cc056355fb9f98fe0acb684

laugmanuel:
@ekohl, currently I'm a little bit hesitant as I can't quite overlook the changes/impact and get a full picture. WDYT about closing this PR in favour of new ones based on your commits (2c2a3223aa1338985cc056355fb9f98fe0acb684 and 58aa20ccded225c5acab3cd2ad24f6e595f6632a)?

ekohl:
I would be OK with that.

ekohl:
I opened https://github.com/voxpupuli/puppet-redis/pull/396

PR #392, opened 2021-02-23
Drop Puppet 5 support

ekohl:


bastelfreak (approved):

ekohl:
Looks like it fails on Debian distros because puppet-bolt isn't available.

kenyon:
> Looks like it fails on Debian distros because puppet-bolt isn't available.

Ah yes, you have to install `https://apt.puppet.com/puppet-release-$facts['os']['distro']['codename'].deb` to enable the puppet-tools component of the repo.

ekohl:
Ah, kind of weird that there's a difference between deb and RPM repos for that.

ehelms:
I noticed that this has a requires on `herculesteam/augeasproviders_core` which currently does not support Puppet 7:

https://github.com/hercules-team/augeasproviders_core/blob/master/metadata.json#L57

bastelfreak:
I pinged Raphael for a new augesproviders_core release

bastelfreak:
@ekohl can you rebase please?

ekohl:
I've rebased this to a minimal version, but we haven't dropped Puppet 5 yet and given it's only metadata now (no actual code), I'd like to consider the alternative where we just add Puppet 7.

bastelfreak (approved):

ekohl:
I've opened https://github.com/voxpupuli/puppet-redis/pull/400 to add Puppet 7 support.

ekohl:
If https://github.com/voxpupuli/puppet-redis/pull/406 is merged, this should also get in.

ekohl:
Rebased

bastelfreak (approved):

PR #395, opened 2021-04-17
camptocamp/systemd: allow 3.x

bastelfreak:
also contains https://github.com/voxpupuli/puppet-redis/pull/394


ekohl (changes requested):
I prefer https://github.com/voxpupuli/puppet-redis/pull/396. A bit more invasive, but much more correct IMHO.

PR #403, opened 2021-06-17
Use Puppet-Datatype Sensitive for $masterauth and $auth_pass

cocker-cc:


kenyon (approved):

ekohl:
Looks like there's a merge conflict now. Would you mind rebasing?

manifests/init.pp
```diff
@@ -331,6 +331,12 @@
   Integer[0] $cluster_migration_barrier                          = 1,
   Hash[String[1], Hash] $instances                               = {},
 ) inherits redis::params {
+  $masterauth_unsensitive = if $masterauth =~ Sensitive {
+    $masterauth.unwrap
+  } else {
+    $masterauth
+  }
```
ekohl, line 338: Where is this used?

ekohl (commented):
Should we also set `show_diff` on the file resource if the value was passed in as `Sensitive`? I think something like this should work:
```puppet
file { '...':
  show_diff => $var !~ Sensitive,
}
```

manifests/init.pp
cocker-cc: Good Question.  It's used only in `manifests/instance.pp`, but not in `manifests/init.pp`.  I will remove these Lines.

cocker-cc:
> Should we also set `show_diff` on the file resource if the value was passed in as `Sensitive`? I think something like this should work:
> 
> ```puppet
> file { '...':
>   show_diff => $var !~ Sensitive,
> }
> ```

With EPP instead of ERB the  `file`-Resource would behave automatically.  I created [this PR](https://github.com/voxpupuli/puppet-redis/pull/407) for therefor.

ekohl:
Needs a rebase now that that PR is merged.

cocker-cc:
> rebase

Done.

templates/redis.conf.epp
```diff
@@ -1,74 +1,74 @@
...
+  Optional[String[1]]                                $syslog_facility,
+  Integer[1]                                         $databases,
+  Boolean                                            $save_db_to_disk,
+  Hash                                               $save_db_to_disk_interval,
+  Boolean                                            $stop_writes_on_bgsave_error,
+  Boolean                                            $rdbcompression,
+  Variant[String[1], Boolean]                        $dbfilename,
+  Stdlib::Absolutepath                               $workdir,
+  Optional[String[1]]                                $slaveof,
+  Optional[Variant[String[1], Sensitive[String[1]]]] $masterauth,
```
ekohl, line 24: I think you're missing the unwrap for this that got lose in a rebase.

cocker-cc: No, in EPP you don't need `unwrap`.  Thats the Point.

ekohl (approved):

ekohl:
Thanks!

PR #405, opened 2021-07-16
set correct redis bin path in systemd for instances when using SCL

alexskr:
redis binary distributed via SCL is located in non-standard location -  /opt/rh/rh-redis5/root/bin/  
this PR updates systemd template to set correct path for redis-server binary when SCL

### DIFF
--- a/templates/service_templates/redis.service.erb
+++ b/templates/service_templates/redis.service.erb
@@ -13,9 +13,9 @@ Type=forking
 ExecStart=/usr/bin/redis-server <%= @redis_file_name %>
 <% else -%>
 Type=notify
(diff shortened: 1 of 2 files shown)


templates/service_templates/redis.service.erb
```diff
@@ -13,9 +13,9 @@ Type=forking
 ExecStart=/usr/bin/redis-server <%= @redis_file_name %>
```
ekohl, line 13: I wondered if this shouldn't also use the variable, but then I realized we should drop 16.04: https://github.com/voxpupuli/puppet-redis/pull/406

ekohl (approved):
This looks correct to me. I guess you ran into this when using `redis::instance`, right? I must admit that I've only used the single instance which uses the service file from the RPM.

PR #407, opened 2021-07-19
Use EPP for Configfile

cocker-cc:
- Use EPP instead of ERB for Configfile, as a Preparation.  EPP is able
  to handle Datatype Sensitive natively.


manifests/instance.pp
```diff
@@ -355,7 +355,81 @@
...
+        protected_mode                => $protected_mode,
+        port                          => $port,
+        tcp_backlog                   => $tcp_backlog,
+        bind_arr                      => $bind_arr,
+        unixsocket                    => $unixsocket,
+        unixsocketperm                => $unixsocketperm,
+        timeout                       => $timeout,
+        tcp_keepalive                 => $tcp_keepalive,
+        log_level                     => $log_level,
+        _real_log_file                => $_real_log_file,
```
ekohl, line 372: Do we need to pass this as `_real_log_file`?
```suggestion
        log_file                => $_real_log_file,
```

templates/redis.conf.epp
```diff
@@ -214,7 +286,7 @@ dir <%= @workdir %>
 #    an error "SYNC with master in progress" to all the kind of commands
 #    but to INFO and SLAVEOF.
 #
-slave-serve-stale-data <% if @slave_serve_stale_data -%>yes<% else -%>no<% end %>
+slave-serve-stale-data <%= $slave_serve_stale_data ? { true => 'yes', default => 'no' } %>
```
ekohl, line 289: I wonder if we should use `bool2str` from stdlib. It is a common pattern here

templates/redis.conf.epp
```diff
@@ -173,7 +245,7 @@ rdbcompression <% if @rdbcompression -%>yes<% else -%>no<% end %>
 rdbchecksum yes
 
 # The filename where to dump the DB
-<% if @dbfilename %>dbfilename <%= @dbfilename %><% else %># dbfilename dump.rdb<% end %>
+<% if $dbfilename { %>dbfilename <%= $dbfilename %><% } else { %># dbfilename dump.rdb<% } %>
```
ekohl, line 248: How about we fix this a layer higher and in config.pp do `pick($dbfilename, 'dump.rdb')`. Then require a `String[1]` as a value.

manifests/instance.pp
```diff
@@ -355,7 +355,81 @@
...
+        syslog_facility               => $syslog_facility,
+        databases                     => $databases,
+        save_db_to_disk               => $save_db_to_disk,
+        save_db_to_disk_interval      => $save_db_to_disk_interval,
+        stop_writes_on_bgsave_error   => $stop_writes_on_bgsave_error,
+        rdbcompression                => $rdbcompression,
+        dbfilename                    => $dbfilename,
+        workdir                       => $workdir,
+        slaveof                       => $slaveof,
+        masterauth_unsensitive        => $masterauth_unsensitive,
```
ekohl, line 383: Looks like you mixed 2 patches:
```
manifests/instance.pp:383:variable_scope:WARNING:top-scope variable being used without an explicit namespace
```
```suggestion
        masterauth_unsensitive        => $masterauth,
```

manifests/instance.pp
cocker-cc: Yes, absolutely.  
Fixed.

templates/redis.conf.epp
cocker-cc: Nice, didn't know that one.  
Implemented

templates/redis.conf.epp
cocker-cc: I guess the Point here is to decide, if this Line should be active or not.  It's not so much the Filename itself.  

In `manifests/instance.pp` we could do something like
```
  $_real_dbfilename = if $dbfilename =~ String {
    $dbfilename
  } elsif $dbfilename = true {
    'dump.rdb'
  } else {
    undef
  }
```

But that would not help that much for the Template to decide between active or inactive Line.

manifests/instance.pp
cocker-cc: Ups, You're right.

cocker-cc:
Compare [the recent Edit](https://github.com/cocker-cc/puppet-redis/compare/30dc253..2dec17b).

templates/redis.conf.epp
cocker-cc: This would also help, that if the User passes the Boolean `true` the resulting Filename would also be `true`.

ekohl: Oh, my apologies; I completely missed the comment bit in `else`. Now it makes sense to have `if/else`.

ekohl (approved):

PR #413, opened 2021-08-29
Honor redis::sentinel::package_ensure

kajinamit:
This change ensures that the package_ensure parameter is properly used in the redis::sentinel class

The package_ensure parameter of the redis_sentinel class is supposed to define status of the redis-sentinel package but in fact it is not used. This fixes the ignored parameter.

Also, default of the parameter is updated following the recent change of redis::package_ensure. This is required to avoid conflicting declaration of redis package caused by a different status; 'installed' vs 'present'.

### DIFF
--- a/spec/classes/redis_sentinel_spec.rb
+++ b/spec/classes/redis_sentinel_spec.rb
@@ -29,6 +29,14 @@ describe 'redis::sentinel' do
         end
       end
 
+      let(:sentinel_package_name) do
+        if facts[:os]['family'] == 'Debian'
+          'redis-sentinel'
+        else
+          'redis'
+        end
+      end
+
       describe 'without parameters' do
         let(:expected_content) do
           <<CONFIG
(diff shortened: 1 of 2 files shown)


root-expert:
Hello @kajinamit, thanks for the contribution!
Can you add a unit test for this?

spec/classes/redis_sentinel_spec.rb
```diff
@@ -112,6 +122,13 @@
         it { is_expected.to compile.with_all_deps }
         it { is_expected.to create_class('redis::sentinel') }
         it { is_expected.to contain_file(config_file_orig).with_content(expected_content) }
+
+        if facts[:os]['family'] == 'Debian'
+          it { is_expected.to contain_package('redis-sentinel').with_ensure('latest') }
+        else
+          it { is_expected.not_to contain_package('redis-sentinel') }
+          it { is_expected.to contain_package('redis').with_ensure('latest') }
+        end
```
smortex, line 131: Ensuring something is not managed is a bit error prone.  I would suggest something like:

```suggestion

        if facts[:os]['family'] == 'Debian'
          sentinel_package_name = 'redis-sentinel'
        else
          sentinel_package_name = 'redis'
        end
        it { is_expected.to contain_package(sentinel_package_name).with_ensure('latest') }
```

kajinamit: My intention was to leave the original test case but I agree with you and that looks truly better. I'll update the change.

smortex (approved):
LGTM

root-expert (approved):

PR #415, opened 2021-08-29
Support Debian 11

root-expert:

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -16,13 +16,19 @@ class redis::params inherits redis::globals {
       $pid_file                  = '/var/run/redis/redis-server.pid'
       $workdir                   = '/var/lib/redis'
       $bin_path                  = '/usr/bin'
-      $daemonize                 = true
+      $daemonize                 = $facts['os']['distro']['codename'] ? {
+        'bullseye' => false,
+        default    => true,
+      }
       $service_name              = 'redis-server'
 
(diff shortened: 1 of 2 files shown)


ekohl:
Looks like it fails on installing bolt because the package is not available. I couldn't find an issue on https://tickets.puppet.com but that may just be my limited JIRA search skills.

root-expert:
> Looks like it fails on installing bolt because the package is not available. I couldn't find an issue on https://tickets.puppet.com but that may just be my limited JIRA search skills.

Yea, there isn't a Puppet Server package as well. Let's give them some time :smile:

ekohl:
My ideal goal for this module was to leverage instances of `redis-server@.service` which is shipped on Debian & Ubuntu. For example on Ubuntu 18.04:

```console
# dpkg -L redis-server | grep .service
/lib/systemd/system/redis-server.service
/lib/systemd/system/redis-server@.service
```

It's not shipped on Debian 9, but we can drop Debian 9 support now that it's only maintained as LTS (as opposed to the Debian project itself).

Then we would get out of the business of managing systemd files. An additional benefit is that you would get the security hardening that's in place.

Sadly on Red Hat-based distros this isn't shipped. That can be solved by shipping a `redis@.service` just for EL though.

The migration may also be complex so I never got around to it, but that's the vision I had. We should probably write it down in an issue.

PR #417, opened 2021-08-30
Skip ensuring package if sentinel requires no additional package

kajinamit:
This change prevents duplicate package management by redis and redis::package in the distros which provide the common package for both Redis and Sentinel.

Currently both of the redis class and the redis::sentinel class have the package_ensure parameter which define status of the packages to install redis and redis sentinel.
However, in some distros like CentOS, the single redis package is enough to install both redis and sentinel, and we should set the same value to these two package_ensure parameters to avoid conflict between two package resources maintained by these two classes.
This change disables the package resource in the redis::sentinel class when the single package is used, to avoid such redundant requirements.

### DIFF
--- a/manifests/sentinel.pp
+++ b/manifests/sentinel.pp
@@ -56,7 +56,8 @@
 #   The name of the package that installs sentinel.
 #
 # @param package_ensure
-#   Do we ensure this package.
+#   Do we ensure this package. This parameter takes effect only if
+#   an independent package is required for sentinel.
 #
 # @param parallel_sync
 #   How many slaves can be reconfigured at the same time to use a
@@ -147,9 +148,11 @@ class redis::sentinel (
 
   require 'redis'
 
-  ensure_packages([$package_name], {
-    ensure => $package_ensure
-  })
+  if $package_name != $redis::package_name {
+    ensure_packages([$package_name], {
+      ensure => $package_ensure
+    })
+  }
   Package[$package_name] -> File[$config_file_orig]
 
   $sentinel_bind_arr = delete_undef_values([$sentinel_bind].flatten)
--- a/spec/classes/redis_sentinel_spec.rb
+++ b/spec/classes/redis_sentinel_spec.rb
@@ -29,6 +29,10 @@ describe 'redis::sentinel' do
         end
       end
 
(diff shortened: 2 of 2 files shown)


ekohl (commented):
Why is this needed? The whole point of `ensure_packages` is that you can call it multiple times.

kajinamit:
The problem is that ensure_packages fails if the ensure parameter has different values.
For example the following succceds
`
package{'ruby':
  ensure => 'present'
}

ensure_packages(['ruby'], {
  ensure => 'latest'
})
`
while the following fails because of duplicate declaration.
`
package{'ruby':
  ensure => 'present'
}

ensure_packages(['ruby'], {
  ensure => 'latest'
})
`

With current implementation you should update both redis::package_ensure and redis::sentinel::package_ensure with the consistent value even though these two results in declaring the same package.

ekohl:
I'm still not sure I really agree with this. If there is a mismatch, then it's silently masking that. If one asked for latest while the other didn't, how do you know what the end result will be?

I wonder if it wouldn't make more sense to make `package_ensure` default to `undef` and then use `pick($package_ensure, $redis::package_ensure)` for the real value. That way the defaults are still sane, but if a user specified it and there is a mismatch, they'll be notified that they asked for something invalid.

kajinamit:
I found that even the current default('installed') doesn't work with older puppetlabs-stdlib 7.1.1 .

```
package{'ruby':
  ensure => 'installed'
}

ensure_packages(['ruby'], {
  ensure => 'installed'
})
```

This results in duplicate declaration with puppetlabs-stdlib 7.1.1. This is because in older  puppetlabs-stdlib, ensure_packages replaces 'installed' by 'present'. The behavior has been changed in 8.0.0 and now ensure_packages internally do the opposite and only setting 'installed' for both can work...

kajinamit:
@ekohl I think that is an option here but I'm afraid that leaves the problem with older puppetlabs-stdlib. If this module supports only puppetlabs-stdlib>=8.0.0 that is an option but to workaround the problem I explained we need to skip declaring the resource.

ekohl (approved):
I guess this is also a valid way of ensuring it. Thanks!

PR #424, opened 2021-10-12
service_template RuntimeDirectory

aamerik:
Set instance specific RuntimeDirectory in systemd unit files when running multiple instances for security/chroot purposes.  Convert redis.service.erb to EPP.


ekohl:
@smortex want to take a last look?

PR #426, opened 2021-11-08
manage additional redis modules

ymartin-ovh:
We want to manage loadmodule directives to allow someone to manage additional redis modules.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -108,6 +108,8 @@
 #   The lag in seconds
 # @param min_slaves_to_write
 #   Minimum number of slaves to be in "online" state
+# @param modules
+#   Additional redis modules to load (.so path)
 # @param no_appendfsync_on_rewrite
 #   If you have latency problems turn this to 'true'. Otherwise leave it as
 # @param notify_keyspace_events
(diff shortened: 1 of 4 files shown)


REFERENCE.md
```diff
@@ -549,6 +550,14 @@ Minimum number of slaves to be in "online" state
 
 Default value: `0`
 
+##### <a name="modules"></a>`modules`
+
+Data type: `Array[Stdlib::Absolutepath]`
+
+Absolute path of additional modules to load
```
smortex, line 557: This string seems outdated comgared to the changes bellow.  Do not bother updating `REFERENCE.md` since it can easily introduce conflicts with other contributors: it will be updated during the release process :wink:

smortex (changes requested):
Thank you!  Can you please add corresponding unit test so that we are sure the feature will not silently broke in the future?

REFERENCE.md
ymartin-ovh: Ok, I will remove this.

spec/acceptance/suites/default/redis_loadmodule.rb
```diff
@@ -0,0 +1,42 @@
+require 'spec_helper_acceptance'
```
root-expert, line 1: Add unit tests instead of acceptance.
Acceptance are a little bit overkill for checking the config IMO

bastelfreak (approved):

PR #427, opened 2021-11-15
Fix issue redis.conf being world readable

### LINKED ISSUE #425: redis.conf is readable by all

- Puppet: all
- Ruby: all
- Distribution: all 
- Module version: 8.2.0

Install a Redis using default conf with a password
```
class{"redis":
    requirepass => "foo",
}
```

## What are you seeing
The file /etc/redis/redis.conf which contains the password is readable by all users. Thus, it is trivial to a local unauthorized user to connect to the protected db.

The file /etc/redis/redis.conf should be readable only by redis user and its group.

## Output log
N/A

I would advocate setting the `config_file_mode` to `640` (see https://github.com/voxpupuli/puppet-redis/blob/3f7ecdd69061be0436c04c1dc9df8c20d884ea5f/manifests/init.pp#L276 )
It has an easy workaround (just set manually the  `config_file_mode` ), however, since I assume a lot of instance are using default, I would think the more secure value should be the default. Plus `640` is the default redis mode for the package.

jnguiot:
Change default mode for redis.conf in order to protect it

Fixes #425

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -275,7 +275,7 @@ class redis (
   Stdlib::Absolutepath $config_dir                               = $redis::params::config_dir,
   Stdlib::Filemode $config_dir_mode                              = $redis::params::config_dir_mode,
   Stdlib::Absolutepath $config_file                              = $redis::params::config_file,
-  Stdlib::Filemode $config_file_mode                             = '0644',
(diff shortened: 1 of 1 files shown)


root-expert (approved):

root-expert:
Thanks for the PR!

PR #435, opened 2022-02-07
Drop default PPA setting and only install EPEL on EL7

ekohl:
In EL8 (and EL9) Redis is included in the OS itself. Only on EL7 it needs to be installed from EPEL. Even then, it can be installed via Software Collections which are included in the OS.

### DIFF
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -1,7 +1,7 @@
 require 'voxpupuli/acceptance/spec_helper_acceptance'
 
 configure_beaker do |host|
-  if fact_on(host, 'osfamily') == 'RedHat'
+  if fact_on(host, 'osfamily') == 'RedHat' && fact_on(host, 'operatingsystemmajrelease') == '7'
     install_module_from_forge_on(host, 'puppet/epel', '>= 3.0.0')
   end
   host.install_package('puppet-bolt')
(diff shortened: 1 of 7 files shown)


ekohl:
I just realized the PPA mentioned hasn't been updated since 2014: https://launchpad.net/~chris-lea/+archive/ubuntu/redis-server

bastelfreak (approved):

spec/spec_helper_acceptance.rb
```diff
@@ -3,9 +3,9 @@
 configure_beaker do |host|
   case fact_on(host, 'operatingsystem')
   when 'CentOS'
-    host.install_package('epel-release')
+    install_module_from_forge_on(host, 'puppet/epel', '>= 3.0.0')
```
bastelfreak, line 6: previously, something complained about the missing upper limit. maybe that's gone now. updating this in the acceptancetests is always anoying

ekohl: The API it's using is `include epel`. It could probably work with `>= 0`.

ekohl:
I think this PR escalated a bit in scope, but it brings everything closer.

For now I left out fixing puppet-strings since some more changes are needed there.

bastelfreak (approved):

ekohl:
I think this should now be ready.

PR #438, opened 2022-02-07
Modernize the acceptance tests

ekohl:
This takes the lessons from https://github.com/rubocop/rubocop-rspec/issues/1231 and applies them here.


spec/acceptance/suites/default/redis_adminstration_spec.rb
```diff
@@ -1,28 +1,30 @@
...
 # systcl settings are untestable in docker
 describe 'redis::administration', unless: default['hypervisor'] =~ %r{docker} do
-  it 'runs successfully' do
-    pp = <<-EOS
-    include redis
-    include redis::administration
-    EOS
+  def execute_with_warning
+    have_attributes(stderr: %r{WARNING})
+  end
```
ekohl, line 9: This here is to combine things with `.and` since you can't use `.not_to().and()`.

bastelfreak (approved):

bastelfreak (approved):

ekohl:
Updated to top the Debian backport acceptance test. See that commit for details.

bastelfreak (approved):

alexjfisher (approved):

ekohl:
I had one copy-paste error where `.not_to` was replaced by `.to`. Fixed that.

spec/acceptance/suites/default/redis_multi_instances_one_host_spec.rb
```diff
@@ -23,67 +23,55 @@
...
+      PUPPET
+    end
   end
 
-  describe service(redis_name) do
-    it { is_expected.not_to be_enabled }
-    it { is_expected.not_to be_running }
-  end
+  specify { expect(package(redis_name)).not_to be_installed }
+  specify { expect(service(redis_name)).not_to be_enabled.and be_running }
```
pirj, line 55: This line confuses me. As mentioned in the comment above - isn't `not_to ...and` disallowed in RSpec in some way?

pirj: Alright, here it is:
```
        `expect(...).not_to matcher.and matcher` is not supported, since it creates a bit of an ambiguity. Instead, define negated versions of whatever matchers you wish to negate with `RSpec::Matchers.define_negated_matcher` and use `expect(...).to matcher.and matcher`.
```

I'd suggest either matching against a negated predicate if it exists (`disabled?` and `stopped?`) with a regular `.to` runner:
```ruby
specify { expect(service(redis_name)).to be_disabled.and be_stopped }
```
If those predicates are not available on what `service` returns, define negated matchers for `be_running` and `be_enabled`.

Or just plain split this expectation in two:
```ruby
specify do
  expect(service(redis_name)).not_to be_enabled
  expect(service(redis_name)).not_to be_running
end
```

Whatever you like the most.

ghoneycutt: Two different tests is much easier to read and maintain.

pirj: I suppose you meant two spec examples.
Sure, I agree, unless expensive setup is involved, which I've heard, is sometimes the case with serverspec.

ekohl: You're absolutely right that this line is vague and confusing.

It is very common to always want it both enabled and running or both disabled and stopped. I've been wondering if it makes sense to have some short hand for it, but I struggled to find a good name for it.

Looking at https://serverspec.org/resource_types.html#service there is no `be_disabled` nor `be_stopped`. I'll think a bit more about this.

ekohl: For now I went with the recommendation to split it in two statements. That was the easiest.

For example, `not_to be_running` is clear. But is `to be_stopped` the same? `to be_stopped` could imply it should be present but stopped while `not_to be_running` doesn't care whether the service is present or not. Splitting it in two avoids these issues.

ekohl:
I think this should be good to merge. I'm leaning to dropping Archlinux from the metadata for now since the tests obviously fail. Also #439 suggests it's entirely broken at the moment.

bastelfreak (approved):
I'm fine with dropping Arch Linux from the metadata.json for now, or merging this as is.

ekohl:
That would be https://github.com/voxpupuli/puppet-redis/pull/441. I still think the changed in this PR that relate to Arch are good to keep in. Those are in separate commits so easy to distinguish.

@pirj do you think this is now all good to merge?

zilchms:
2 year ping @ekohl :) do you want to rebase this to the current master?

ekohl:
Rebased to resolve merge conflicts. I didn't run it locally yet, so it may have broken something in the rebase.

spec/acceptance/suites/default/redis_sentinel_one_node_spec.rb
```diff
@@ -3,54 +3,38 @@
...
-    }
-    EOS
-
-    apply_manifest(pp, catch_failures: true)
-    apply_manifest(pp, catch_changes: true)
-  end
-
-  describe package(redis_name) do
-    it { is_expected.to be_installed }
+  redis_name = fact('osfamily') == 'Debian' ? 'redis-server' : 'redis'
```
zilchms, line 6: osfamily is a legacy fact; but this might be alright of we have more legacy facts that we all fix in one PR

spec/acceptance/suites/default/redis_spec.rb
```diff
@@ -3,21 +3,10 @@
...
-                 'redis-server'
-               else
-                 'redis'
-               end
-
-  it 'runs successfully' do
-    pp = <<-EOS
-    include redis
-    EOS
+  redis_name = fact('osfamily') == 'Debian' ? 'redis-server' : 'redis'
```
zilchms, line 6: same legacy fact here

zilchms (approved):
I dont know if we want to fix the legacy facts in this PR or do all legacy stuff in a second one. Additionally we should look at why the Puppet 7 on Centos 8 test times out.
Otherwise this looks good :)

ekohl:
> Additionally we should look at why the Puppet 7 on Centos 8 test times out.

It's common for the CentOS 8 tests to time out. I suspect pulling from quay.io may cause problems.

ekohl:
Updated to drop legacy facts in the test suite.

PR #440, opened 2022-02-22
Add Debian 11 support

ekohl:
Right now this is untested. I'll have to see what CI says.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -36,7 +36,8 @@
       "operatingsystem": "Debian",
       "operatingsystemrelease": [
         "9",
-        "10"
+        "10",
+        "11"
       ]
     },
     {
--- a/spec/classes/redis_spec.rb
+++ b/spec/classes/redis_spec.rb
@@ -1426,7 +1426,7 @@ describe 'redis' do
             |RuntimeDirectory=#{service_name}
             |RuntimeDirectoryMode=2755
             |Type=notify
-            |ExecStart=/usr/bin/redis-server #{config_file} --supervised systemd
+            |ExecStart=/usr/bin/redis-server #{config_file} --supervised systemd --daemonize no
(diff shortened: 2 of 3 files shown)


ekohl:
It looks like multi server doesn't work properly:
```
   Feb 22 10:39:09 debian11-64.example.com systemd[1]: redis-server-6379.service: Got notification message from PID 2027, but reception only permitted for main PID which is currently not known
```

Haven't looked into this further.

root-expert:
If you need some insight https://github.com/voxpupuli/puppet-redis/pull/415. Although I haven't done much work either

ekohl:
Ah, my bad. I somehow missed that one.

bastelfreak (approved):

ekohl:
This is still a draft, no need to approve just yet.

ekohl:
This should now be ready.

root-expert:
Is it okay to change that for all OSes?
Maybe it should be configurable only for Debian 11?

ekohl:
I thought about that, but we set `Type=notify` which means it can't daemonize. That was precisely the problem why it wouldn't start. So just like setting the supervisor to systemd (which must also be true with `Type=notify`), this must also always be true. And to me that CI passes with it says it's safe enough.

ekohl:
Oh, and ideally speaking we would actually use the shipped `redis@.service` included in the Debian package since that actually contains more hardening.

root-expert:
> I thought about that, but we set `Type=notify` which means it can't daemonize. That was precisely the problem why it wouldn't start. So just like setting the supervisor to systemd (which must also be true with `Type=notify`), this must also always be true. And to me that CI passes with it says it's safe enough.

You're right I had noted that on my PR as well

root-expert (approved):

PR #443, opened 2022-03-24
Add tls settings to the redis::sentinel class and the redis-sentinel.conf template

tparkercbn:
Added tls settings to the redis sentinel submodule.  If you have tls only enabled for redis not having tls on the sentinels makes them useless.

### DIFF
--- a/manifests/sentinel.pp
+++ b/manifests/sentinel.pp
@@ -70,12 +70,22 @@
 #   Number of sentinels that must agree that a master is down to
 #   signal sdown state.
 #
+# @param sentinel_announce_hostnames
+#   Whether or not sentinels will announce hostnames instead of ip addresses
+#   to clients.  This can be required for TLS.
+#
 # @param sentinel_bind
 #   Allow optional sentinel server ip binding.  Can help overcome
 #   issues arising from protect-mode added Redis 3.2
 #
 # @param sentinel_port
 #   The port of sentinel server.
+#
+# @param sentinel_resolve_hostnames
+#   Whether or not sentinels can resolve hostnames to ip addresses.
+#
+# @param sentinel_tls_port
+#   Configure which TLS port to listen on.
 #
 # @param service_group
 #   The group of the config file.
@@ -88,6 +98,24 @@
 #
 # @param service_enable
 #   Enable the service at boot time.
+#
+# @param tls_cert_file
+#   Specify which X.509 certificate file to use for TLS connections.
+#
+# @param tls_key_file
+#   Specify which privaye key file to use for TLS connections.
+#
+# @param tls_ca_cert_file
+#   Specify which X.509 CA certificate(s) bundle file to use.
+#
+# @param tls_ca_cert_dir
+#   Specify which X.509 CA certificate(s) bundle directory to use.
+#
+# @param tls_auth_clients
+#   Specify if clients and replicas are required to authenticate using valid client side certificates.
+#
+# @param tls_replication
+#   Specify if TLS should be enabled on replication links.
 #
 # @param working_dir
 #   The directory into which sentinel will change to avoid mount
@@ -129,13 +157,22 @@ class redis::sentinel (
   Integer[0] $parallel_sync = 1,
   Stdlib::Absolutepath $pid_file = $redis::params::sentinel_pid_file,
   Integer[1] $quorum = 2,
+  Optional[Enum['yes', 'no']] $sentinel_announce_hostnames = undef,
   Variant[Undef, Stdlib::IP::Address, Array[Stdlib::IP::Address]] $sentinel_bind = undef,
   Stdlib::Port $sentinel_port = 26379,
+  Optional[Enum['yes', 'no']] $sentinel_resolve_hostnames = undef,
+  Optional[Stdlib::Port::Unprivileged] $sentinel_tls_port = undef,
   String[1] $service_group = 'redis',
   String[1] $service_name = $redis::params::sentinel_service_name,
   Stdlib::Ensure::Service $service_ensure = 'running',
   Boolean $service_enable = true,
   String[1] $service_user = 'redis',
+  Optional[Stdlib::Absolutepath] $tls_cert_file = undef,
+  Optional[Stdlib::Absolutepath] $tls_key_file = undef,
+  Optional[Stdlib::Absolutepath] $tls_ca_cert_file = undef,
+  Optional[Stdlib::Absolutepath] $tls_ca_cert_dir = undef,
+  Enum['yes', 'no', 'optional'] $tls_auth_clients = 'no',
+  Boolean $tls_replication = false,
   Stdlib::Absolutepath $working_dir = $redis::params::sentinel_working_dir,
   Optional[Stdlib::Absolutepath] $notification_script = undef,
   Optional[Stdlib::Absolutepath] $client_reconfig_script = undef,
--- a/spec/classes/redis_sentinel_spec.rb
+++ b/spec/classes/redis_sentinel_spec.rb
@@ -94,6 +94,7 @@ CONFIG
 
         let(:params) do
           {
+            sentinel_tls_port: 26_380,
             auth_pass: 'password',
             sentinel_bind: '192.0.2.10',
             protected_mode: false,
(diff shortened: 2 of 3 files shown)


metadata.json
```diff
@@ -4,7 +4,7 @@
   "author": "Vox Pupuli",
   "summary": "Redis module",
   "license": "Apache-2.0",
-  "source": "https://github.com/voxpupuli/puppet-redis.git",
+  "source": "https://github.com/cbntss/puppet-redis.git",
```
root-expert, line 7: Oops 😄

root-expert (commented):
Thanks for the PR!
Please add some unit tests for the new parameters 😄

tparkercbn:
Thanks @root-expert is there somewhere I can get a pointer on writing unit tests.  That's where my puppet knowledge falls down.

root-expert:
The unit tests should be added in the [redis_sentinel_spec](https://github.com/voxpupuli/puppet-redis/blob/master/spec/classes/redis_sentinel_spec.rb)  file.

You can add your new parameters with a random value in [this](https://github.com/voxpupuli/puppet-redis/blob/master/spec/classes/redis_sentinel_spec.rb#L97-L107) block and check that the random value is rendered correctly in [this](https://github.com/voxpupuli/puppet-redis/blob/master/spec/classes/redis_sentinel_spec.rb#L112-L130) block.

For more in depth knowledge for testing puppet you can check [rspec-puppet.com](https://rspec-puppet.com/) and our [CONTRIBUTING.md](https://github.com/voxpupuli/puppet-redis/blob/master/.github/CONTRIBUTING.md) for instructions on how to run unit tests. (You can always just push and let our CI do the hard job 😄 )

tparkercbn:
Perfect!  Will do.  Might be on Monday but it looks straightforward.

spec/classes/redis_sentinel_spec.rb
```diff
@@ -94,6 +94,7 @@ class { 'redis':
 
         let(:params) do
           {
+            tls_port: 26380
```
root-expert, line 97: Forgot the comma here

tparkercbn: Yes, yes I did :)  

Thanks!

manifests/sentinel.pp
```diff
@@ -129,13 +157,22 @@
   Integer[0] $parallel_sync = 1,
   Stdlib::Absolutepath $pid_file = $redis::params::sentinel_pid_file,
   Integer[1] $quorum = 2,
+  Optional[Boolean] $sentinel_announce_hostnames = undef,
```
root-expert, line 160: This and `sentinel announce-hostnames` should be an `Enum['yes', 'no']` instead of Boolean that converts to yes or no.

Optional can remain I think

tparkercbn: What is the standard you are using?  The rest of the code seems to use Boolean when it's yes, no unless there is another possible value.

tparkercbn: For example: 
protected-mode <%= @protected_mode ? 'yes' : 'no' %>

root-expert: We try to use Enum instead of booleans convering to some values

https://voxpupuli.org/docs/reviewing_pr/

> Does a new param map to an option in a config file of a service? The Parameter should accept the possible values that the service allows. For example ‘on’ and ‘off’. Don’t accept a boolean that will be converted to ‘on’ or ‘off’

tparkercbn: okay.  I will fix.  I was just trying to keep things consistent with what was already there.

tparkercbn:
There is an odd CI failure for the port number being too long too.

root-expert (approved):

manifests/sentinel.pp
```diff
@@ -129,13 +157,22 @@
   Integer[0] $parallel_sync = 1,
   Stdlib::Absolutepath $pid_file = $redis::params::sentinel_pid_file,
   Integer[1] $quorum = 2,
+  Optional[Enum['yes', 'no']] $sentinel_announce_hostnames = undef,
   Variant[Undef, Stdlib::IP::Address, Array[Stdlib::IP::Address]] $sentinel_bind = undef,
   Stdlib::Port $sentinel_port = 26379,
+  Optional[Enum['yes', 'no']] $sentinel_resolve_hostnames = undef,
+  Optional[Stdlib::Port] $sentinel_tls_port = undef,
```
bastelfreak, line 164: should we use https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/types/port/unprivileged.pp ? I guess noone wants to run this as root on a lowlevel port?

tparkercbn:
I changed tls_replication to use the Enum['yes', 'no'] but that makes it not the same as the identical variable in the redis class.  Which way do you prefer?

manifests/sentinel.pp
tparkercbn: I really doubt anyone would (or should).  I'm good with that change.  Again, mostly trying to be consistent with existing code with many of my choices.  I would do things very differently in this module myself (including using an epp).

root-expert: We use EPPs instead of ERBs as well. If you have some spare time to work on the conversion I'll be more than happy to review it :smile:

tparkercbn:
> I changed tls_replication to use the Enum['yes', 'no'] but that makes it not the same as the identical variable in the redis class. Which way do you prefer?

I think I am going to put this back to a boolean. It would be very confusing the people to have the same variable as a Boolean for one module and an Enum for the submodule.

root-expert:
> > I changed tls_replication to use the Enum['yes', 'no'] but that makes it not the same as the identical variable in the redis class. Which way do you prefer?
> 
> I think I am going to put this back to a boolean. It would be very confusing the people to have the same variable as a Boolean for one module and an Enum for the submodule.

I believe the new options should follow Enum style, but I agree it can be confusing. 
@bastelfreak any opinion here? :smile:

manifests/sentinel.pp
```diff
@@ -122,20 +150,29 @@
   Stdlib::Absolutepath $log_file = $redis::params::sentinel_log_file,
   String[1] $master_name  = 'mymaster',
   Stdlib::Host $redis_host = '127.0.0.1',
-  Stdlib::Port $redis_port = 6379,
+  Stdlib::Port::Unprivileged $redis_port = 6379,
```
root-expert, line 153: I would consider this change (and `sentinel_port` parameter) backwards incompatible, so it's better to stay out of this PR.

tparkercbn:
I have been thinking about the default values for the tls_ options in the sentinel class.  Does it make sense to default them to the values from the redis class?  

For most instalations they should be in sync all the time.

That would mean that the cert paths, and replcation values default to undef still but pulled from redis::tls_*

tparkercbn:
> > > I have been thinking about the default values for the tls_ options in the sentinel class. Does it make sense to default them to the values from the redis class?
> > > For most instalations they should be in sync all the time.
> > > That would mean that the cert paths, and replcation values default to undef still but pulled from redis::tls_*
> > 
> > 
> > This works in puppet on my servers, however doesn't pass the spec tests because the $redis:: variables are not defined. Is there a way to define a more complete precondition so that we can consume those values? The redis class is required in the sentinel code so there will never be a case where we can have the redis::sentinel without redis already being in the catalog.
> 
> Take a look [here](https://rspec-puppet.com/documentation/classes/#specifying-code-to-include-before)

Thanks!  I was confused because it already had that set for the custom params.  I hadn't noticed it was failing for the default params case.

tparkercbn:
> > > > I have been thinking about the default values for the tls_ options in the sentinel class. Does it make sense to default them to the values from the redis class?
> > > > For most instalations they should be in sync all the time.
> > > > That would mean that the cert paths, and replcation values default to undef still but pulled from redis::tls_*
> > > 
> > > 
> > > This works in puppet on my servers, however doesn't pass the spec tests because the $redis:: variables are not defined. Is there a way to define a more complete precondition so that we can consume those values? The redis class is required in the sentinel code so there will never be a case where we can have the redis::sentinel without redis already being in the catalog.
> > 
> > 
> > Take a look [here](https://rspec-puppet.com/documentation/classes/#specifying-code-to-include-before)
> 
> Thanks! I was confused because it already had that set for the custom params. I hadn't noticed it was failing for the default params case.

I am still not able to get the spec tests to work.  Have I missed something?

root-expert:
Let's use plain defaults @tparkercbn :smile:

root-expert:
Let's use plain defaults @tparkercbn :smile:

tparkercbn:
> Let's use plain defaults @tparkercbn smile

Yep.  That's what I think too.  

I am going to keep the tls_replication setting a Boolean and tls_auth_clients with the default to 'no' so that the types are consistent between the main and sentinel classes.

root-expert (approved):
Thank for the contribution!

tparkercbn:
> Thank for the contribution!

Thank you for all your help! :)

PR #445, opened 2022-04-06
Add support for Redis6 features

fraenki:

This is basically a partial re-submission of PR #388 to add Redis 6 options.

However, there are some notable differences to the original submission:

* this time with backwards compatibility in mind
* not duplicating the redis.conf template and tests
* and without any refactoring (to keep the diff as small as possible)

### DIFF
--- a/spec/classes/redis_spec.rb
+++ b/spec/classes/redis_spec.rb
@@ -576,13 +576,13 @@ describe 'redis' do
       describe 'with parameter maxmemory_policy' do
         let(:params) do
           {
-            maxmemory_policy: '_VALUE_'
+            maxmemory_policy: 'noeviction'
           }
         end
 
         it {
           is_expected.to contain_file(config_file_orig).with(
-            'content' => %r{maxmemory-policy.*_VALUE_}
+            'content' => %r{maxmemory-policy.*noeviction}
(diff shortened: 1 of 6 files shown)


spec/classes/redis_spec.rb
```diff
@@ -1478,6 +1478,146 @@ class { 'redis::globals':
...
+          {
+            rdb_save_incremental_fsync: true,
+          }
+        end
+
+        it {
+          is_expected.to contain_file(config_file_orig).with(
+            'content' => %r{^rdb-save-incremental-fsync yes$}
+          )
+        }
```
bastelfreak, line 1619: You do not have to reformat this but I think having this in a single line is best practice:
```suggestion
        it { is_expected.to contain_file(config_file_orig).with('content' => %r{^rdb-save-incremental-fsync yes$}) }
```

fraenki: Fixed in most recent commit. Thanks :)

bastelfreak (approved):
looks good to me. thanks for the PR! I will keep this open for a few days in case someone else want's to review this as well (I'm not a redis expert).

root-expert (approved):

PR #450, opened 2022-09-27
Add support for managing Redis dnfmodule on EL 8+

ehelms:

### DIFF
--- a/manifests/dnfmodule.pp
+++ b/manifests/dnfmodule.pp
@@ -1,0 +1,18 @@
+# @summary Manage the DNF module
+#
+# On EL8+ and Fedora DNF can manage modules. This is a method of providing
+# multiple versions on the same OS. Only one DNF module stream can be active at the
+# same time.
+#
+# @api private
+class redis::dnfmodule (
+  String[1] $ensure = 'installed',
+  String[1] $module = 'redis',
+) {
+  package { 'redis dnf module':
+    ensure      => $ensure,
+    name        => $module,
+    enable_only => true,
+    provider    => 'dnfmodule',
+  }
+}
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -1,7 +1,7 @@
 # @summary Set a global config for Redis
 #
 # @param scl
-#   Use a specific Software CoLlection on Red Hat based systems
+#   Use a specific Software Collection on Red Hat 7 based systems
 class redis::globals (
   Optional[String] $scl = undef,
 ) {
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -302,7 +302,9 @@
 # @param rdb_save_incremental_fsync
 #   When redis saves RDB file, if the following option is enabled
 #   the file will be fsync-ed every 32 MB of data generated.
-
+# @param dnf_module_stream
+#   Manage the DNF module and set the version. This only makes sense on distributions
+#   that use DNF package manager, such as EL8 or Fedora.
 class redis (
   Boolean $activerehashing                                       = true,
   Boolean $aof_load_truncated                                    = true,
@@ -433,6 +435,7 @@ class redis (
   Integer[1] $active_defrag_max_scan_fields                      = 1000,
   Optional[Boolean] $jemalloc_bg_thread                          = undef,
   Optional[Boolean] $rdb_save_incremental_fsync                  = undef,
+  Optional[String[1]] $dnf_module_stream                         = undef,
 ) inherits redis::params {
   contain redis::preinstall
   contain redis::install
--- a/spec/acceptance/suites/dnfmodule/redis_spec.rb
+++ b/spec/acceptance/suites/dnfmodule/redis_spec.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+require 'spec_helper_acceptance'
+
+describe 'redis', if: %w[centos redhat].include?(os[:family]) && os[:release].to_i == 8 do
+  redis_name = 'redis'
+
+  it 'runs successfully' do
+    pp = <<-PUPPET
+      class { 'redis':
+        dnf_module_stream => '6',
+        package_ensure    => 'latest',
+      }
+    PUPPET
+
+    # Apply twice to ensure no errors the second time.
+    apply_manifest(pp, catch_failures: true)
+    apply_manifest(pp, catch_changes: true)
+  end
+
+  describe package(redis_name) do
+    it { is_expected.to be_installed }
+  end
+
+  describe service(redis_name) do
+    it { is_expected.to be_running }
+  end
+
+  context 'redis should respond to ping command' do
+    describe command('redis-cli ping') do
+      its(:stdout) { is_expected.to match %r{PONG} }
+    end
+  end
+
+  describe command('redis-server -v') do
+    its(:stdout) { is_expected.to match %r{v=6.+} }
+  end
+end
--- a/spec/classes/redis_spec.rb
+++ b/spec/classes/redis_spec.rb
@@ -81,6 +81,21 @@ describe 'redis' do
             end
           end
         end
+
+        describe 'with manage_dnf_module true', if: facts[:os]['family'] == 'RedHat' && facts[:os]['release']['major'].to_i == 8 do
+          let(:pre_condition) do
+            <<-PUPPET
+            class { 'redis':
+              manage_package    => true,
+              dnf_module_stream => '6',
+            }
+            PUPPET
+          end
+
+          it { is_expected.to compile.with_all_deps }
+          it { is_expected.to contain_package('redis dnf module').with_ensure('6').that_comes_before('Package[redis]') }
+          it { is_expected.to contain_package('redis').with_name('redis') }
+        end
       end
 
       context 'with managed_by_cluster_manager true' do
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -6,4 +6,15 @@ class redis::install {
       ensure => $redis::package_ensure,
     }
   }
+
+  if $redis::dnf_module_stream {
+    if $facts['os']['family'] != 'RedHat' {
+      fail('DNF modules are only supported on the Red Hat OS family')
+    }
+
+    class { 'redis::dnfmodule':
+      ensure => $redis::dnf_module_stream,
+      before => Package[$redis::package_name],
+    }
+  }
 }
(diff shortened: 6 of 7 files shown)


spec/classes/redis_spec.rb
```diff
@@ -81,6 +81,21 @@ class { 'redis::globals':
             end
           end
         end
+
+        describe 'with manage_dnf_module true', if: facts[:os]['family'] == 'RedHat' && facts[:os]['release']['major'].to_i > 7 do
```
ekohl, line 85: I'm never a fan of `> x`
```suggestion
        describe 'with manage_dnf_module true', if: facts[:os]['family'] == 'RedHat' && facts[:os]['release']['major'].to_i >= 8 do
```

manifests/globals.pp
```diff
@@ -2,10 +2,26 @@
...
 # @param scl
 #   Use a specific Software CoLlection on Red Hat based systems
+#
+# @param version
+#   The version of Redis to install and manage.
+#
+# @param manage_dnf_module
+#   Manage the DNF module. This only makes sense on distributions that use DNF
+#   package manager, such as EL8 or Fedora. It also requires Puppet 5.5.20+ or
+#   Puppet 6.15.0+ since they ship the dnfmodule provider.
```
ekohl, line 12: I'd be happy with changing `metadata.json` to 6.15.0+:
https://github.com/voxpupuli/puppet-redis/blob/38b0d2d4bb2abc0806e9f148ce385ae9a28deb38/metadata.json#L28-L33

manifests/globals.pp
```diff
@@ -2,10 +2,26 @@
...
 class redis::globals (
   Optional[String] $scl = undef,
+  $version = '6',
```
ekohl, line 15: It may be confusing because the user can also pass in a version via `package_ensure`. There is also the `redis_server_version` fact (https://github.com/voxpupuli/puppet-redis/blob/master/lib/facter/redis_server_version.rb).

Should this be more focused and instead be `Optional[String[1]] $dnf_module_stream = undef` (and drop the `manage_dnf_module` parameter)?

manifests/dnfmodule.pp
```diff
@@ -0,0 +1,20 @@
+# @summary Manage the DNF module
+#
+# On EL8 and Fedora DNF can manage modules. This is a method of providing
+# multiple versions on the same OS. Only one DNF module can be active at the
```
ekohl, line 4:
```suggestion
# multiple versions on the same OS. Only one DNF module stream can be active at the
```

spec/classes/redis_spec.rb
```diff
@@ -81,6 +81,21 @@ class { 'redis::globals':
...
+            class { 'redis::globals':
+              manage_dnf_module => true,
+              version           => '7',
+            }
+            PUPPET
+          end
+
+          it { is_expected.to compile.with_all_deps }
+
+          it { is_expected.to contain_package('redis').with_name('redis') }
```
ekohl, line 97: Could you add `.that_requires('Package[redis dnf module]')` to this line?

spec/classes/redis_spec.rb
```diff
@@ -81,6 +81,21 @@ class { 'redis::globals':
...
+            <<-PUPPET
+            class { 'redis::globals':
+              manage_dnf_module => true,
+              version           => '7',
+            }
+            PUPPET
+          end
+
+          it { is_expected.to compile.with_all_deps }
+
```
ekohl, line 96: Could you add the expectation `contain_package('redis dnf module').with_ensure('7')`

manifests/globals.pp
```diff
@@ -2,10 +2,26 @@
 #
 # @param scl
 #   Use a specific Software CoLlection on Red Hat based systems
```
ekohl, line 4:
```suggestion
#   Use a specific Software Collection on Red Hat 7 based systems
```

ekohl (changes requested):
It would be good to verify the version really is 7 in an acceptance test. I recall calling facter on the command line while using custom Puppet facts was tricky so perhaps you can parse the output of `redis-server -v` like the fact does. It's rather trivial anyway and I suspect some of the code never triggers anymore.

ehelms:
I can't seem to see what I am missing to make the requires relationship work in testing:

```
  1) redis on centos-8-x86_64 without parameters with manage_dnf_module true is expected to contain Package[redis dnf module] that comes before Package[redis]
     Failure/Error: it { is_expected.to contain_package('redis dnf module').with_ensure('6').that_comes_before('Package[redis]') }
       expected that the catalogue would contain Package[redis dnf module] with that comes before Package[redis]
     # ./spec/classes/redis_spec.rb:98:in `block (6 levels) in <top (required)>'

  2) redis on centos-9-x86_64 without parameters with manage_dnf_module true is expected to contain Package[redis dnf module] that comes before Package[redis]
     Failure/Error: it { is_expected.to contain_package('redis dnf module').with_ensure('6').that_comes_before('Package[redis]') }
       expected that the catalogue would contain Package[redis dnf module] with that comes before Package[redis]
     # ./spec/classes/redis_spec.rb:98:in `block (6 levels) in <top (required)>'

  3) redis on redhat-8-x86_64 without parameters with manage_dnf_module true is expected to contain Package[redis dnf module] that comes before Package[redis]
     Failure/Error: it { is_expected.to contain_package('redis dnf module').with_ensure('6').that_comes_before('Package[redis]') }
       expected that the catalogue would contain Package[redis dnf module] with that comes before Package[redis]
     # ./spec/classes/redis_spec.rb:98:in `block (6 levels) in <top (required)>'

  4) redis on redhat-9-x86_64 without parameters with manage_dnf_module true is expected to contain Package[redis dnf module] that comes before Package[redis]
     Failure/Error: it { is_expected.to contain_package('redis dnf module').with_ensure('6').that_comes_before('Package[redis]') }
       expected that the catalogue would contain Package[redis dnf module] with that comes before Package[redis]
```

spec/acceptance/suites/dnfmodule/redis_spec.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+require 'spec_helper_acceptance'
+
+describe 'redis', if: %w[centos redhat].include?(os[:family]) && os[:release].to_i >= 8 do
+  redis_name = 'redis'
+
+  it 'runs successfully' do
```
ekohl, line 8: Looks like CentOS 9 doesn't have a Redis 6 module (Redis 6.2.7 is a regular package).

spec/classes/redis_spec.rb
```diff
@@ -81,6 +81,23 @@ class { 'redis::globals':
...
+              dnf_module_stream => '6',
+            }
+            class { 'redis':
+              manage_package => true,
+            }
+            PUPPET
+          end
+
+          it { is_expected.to compile.with_all_deps }
+          it { is_expected.to contain_package('redis dnf module').with_ensure('6').that_comes_before('Package[redis]') }
```
ekohl, line 98: I suspect `Package<|tag == 'puppet-redis'|>` doesn't match the package. A simpler implementation may be to inline `redis::dnfmodule` in `redis::install` and use a regular relation rather than a collector. In that case you can also consider a regular class parameter in `init.pp` rather than one on globals. With SCL globals makes sense (since it affects paths, service names etc) while with DNF modules it's a lot simpler.

ehelms: I took inspiration from the implementation [here[(https://github.com/puppetlabs/puppetlabs-postgresql/commit/8d0efec9fb5a5df42ed64375c71a11d8b6f90b4c) -- do you recall what inspired that or if your thinking around this has simply change to a more standard model?

ekohl: Mostly because the PostgreSQL module is a lot more complex. For example, if you use the https://www.postgresql.org/ RPMs they're packaged very differently from Red Hat's RPMs. Like Debian's default packages, they are co-installable. That means the version comes back in paths like the data directory. As always, there's upsides and downsides to this.

ehelms: I have changed it around to implement this strategy.

manifests/init.pp
```diff
@@ -302,7 +302,10 @@
 # @param rdb_save_incremental_fsync
 #   When redis saves RDB file, if the following option is enabled
 #   the file will be fsync-ed every 32 MB of data generated.
-
+# @param dnf_module_stream
+#   Manage the DNF module and set the version. This only makes sense on distributions
+#   that use DNF package manager, such as EL8 or Fedora. It also requires Puppet 5.5.20+ or
+#   Puppet 6.15.0+ since they ship the dnfmodule provider.
```
ekohl, line 308: Given the minimum version is specified in `metadata.json`, I'd leave it out here
```suggestion
#   that use DNF package manager, such as EL8 or Fedora.
```

spec/acceptance/suites/dnfmodule/redis_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+require 'spec_helper_acceptance'
+
+describe 'redis', if: %w[centos redhat].include?(os[:family]) && os[:release].to_i == 8 do
+  redis_name = 'redis'
+
+  it 'runs successfully' do
+    pp = <<-PUPPET
+      class { 'redis':
+        manage_repo       => true,
```
ekohl, line 11: Why is this needed?

spec/acceptance/suites/dnfmodule/redis_spec.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+require 'spec_helper_acceptance'
+
+describe 'redis', if: %w[centos redhat].include?(os[:family]) && os[:release].to_i == 8 do
+  redis_name = 'redis'
+
+  it 'runs successfully' do
+    pp = <<-PUPPET
+      class { 'redis':
```
ekohl, line 10: I suspect it runs after a test which has already installed Redis and then the ensure doesn't match. I suspect you can use `package_ensure` to upgrade.
```suggestion
      class { 'redis':
        package_ensure => 'latest',
```

ekohl (changes requested):
The two unit test failures also happen on master so they're unrelated. Acceptance tests do appear related. See inline comment about `package_ensure`.

ehelms:
To address the failing unit tests that are unrelated -- https://github.com/voxpupuli/puppet-redis/pull/452

bastelfreak (approved):

ekohl (approved):

PR #456, opened 2022-12-01
Add support for managing Redis APT repo

zaben903:
Adds support for managing the official Redis [repo](https://redis.io/docs/getting-started/installation/install-redis-on-linux/) on Debian based systems to allow installation of additional Redis versions.

n/a

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -378,7 +378,7 @@ class redis (
   Boolean $redis_apt_repo                                        = false,
   Stdlib::HTTPSUrl $apt_location                                 = 'https://packages.redis.io/deb/',
   String[1] $apt_repos                                           = 'main',
-  Optional[String] $apt_release                                  = $facts.dig('os', 'distro', 'codename'),
+  Optional[String] $apt_release                                  = undef,
   String[1] $apt_key_id                                          = '54318FA4052D1E61A6B6F7BB5F4349D6BF53AA0C',
   String[1] $apt_key_server                                      = 'hkp://keyserver.ubuntu.com/',
   Boolean $rdbcompression                                        = true,


REFERENCE.md
```diff
@@ -649,6 +654,46 @@ Specify upstream (Ubuntu) PPA entry.
 
 Default value: `undef`
 
+##### <a name="-redis--apt_location"></a>`apt_location`
+
+Data type: `Optional[String]`
+
+Specify the URL of the apt repository.
+
+Default value: `https://packages.redis.io/deb/`
```
ekohl, line 663: This doesn't look like the real default. Normally we generate the reference using `rake strings:generate:reference`

ekohl (changes requested):
IMHO it shouldn't use the Redis apt repo by default. By default it should still use the OS repos.

manifests/init.pp
```diff
@@ -363,6 +375,12 @@
   Stdlib::Port $port                                             = 6379,
   Boolean $protected_mode                                        = true,
   Optional[String] $ppa_repo                                     = undef,
+  Boolean $redis_apt_repo                                        = false,
```
bastelfreak, line 378: please add the data here and not in params.pp. that enables puppet-strings to pick it up and put it into the reference.md

manifests/params.pp
```diff
@@ -25,6 +25,12 @@
       $sentinel_log_file         = '/var/log/redis/redis-sentinel.log'
       $sentinel_working_dir      = '/var/lib/redis'
 
+      $apt_location              = 'https://packages.redis.io/deb/'
+      $apt_repos                 = 'main'
+      $apt_release               = $facts.dig('os', 'distro', 'codename')
```
ekohl, line 30: This is the same value as the default in `apt::source` so I'd keep it `undef` by default rather than trying to duplicate it.

spec/spec_helper_acceptance.rb
```diff
@@ -12,4 +12,12 @@
     # https://tickets.puppetlabs.com/browse/MODULES-11275
     host.install_package('puppet-bolt')
   end
+
+  if fact_on(host, 'osfamily') == 'Debian'
+    # APT required for Debian based systems where `$redis::manage_repo` is `true`
+    install_module_from_forge_on(host, 'puppetlabs/apt', '>= 9.0.0')
+
+    # Install prerequisite for Debian based systems
+    host.install_package('lsb-release')
```
ekohl, line 21: Only on Facter 3. Facter 4 can extract it from `/etc/os-release`.

bastelfreak (approved):

PR #474, opened 2023-05-29
Migrate acceptance tests to vp-acceptance v2

traylenator:
Migrate acceptance tests to vp-acceptance v2


traylenator:
The bolt tasks fail because:

```
bolt --modulepath /etc/puppetlabs/code/modules task show
```
is called and not an environment specific path , bolt must be environment aware....

PR #472, opened 2023-05-12
modulesync 6.0.0

bastelfreak:
modulesync 6.0.0

PR #475, opened 2023-05-29
Specify environment module path for bolt

traylenator:

Somewhere between voxpupuli-acceptance last v1 and v2 the modules are installed in a path where bolt can't find them.

Fixes #474 
Fixes #472

### DIFF
--- a/spec/acceptance/redis_cli_task_spec.rb
+++ b/spec/acceptance/redis_cli_task_spec.rb
@@ -4,7 +4,7 @@ require 'spec_helper_acceptance'
 
 describe 'redis-cli task' do
   subject do
-    on(default, "bolt task run --modulepath /etc/puppetlabs/code/modules --targets localhost #{task_name} #{params}", acceptable_exit_codes: [0, 1]).stdout
+    on(default, "bolt task run --modulepath /etc/puppetlabs/code/environments/production/modules --targets localhost #{task_name} #{params}", acceptable_exit_codes: [0, 1]).stdout
   end
 
   let(:task_name) { 'redis::redis_cli' }
--- a/spec/acceptance/suites/scl/redis5_spec.rb
+++ b/spec/acceptance/suites/scl/redis5_spec.rb
@@ -4,11 +4,11 @@ require 'spec_helper_acceptance'
 
 describe 'redis', if: %w[centos redhat].include?(os[:family]) && os[:release].to_i == 7 do
   before(:all) do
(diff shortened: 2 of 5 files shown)


traylenator:
Has to be be a better way of getting the environment than hard-coding it.

spec/acceptance/redis_cli_task_spec.rb
```diff
@@ -4,7 +4,7 @@
 
 describe 'redis-cli task' do
   subject do
-    on(default, "bolt task run --modulepath /etc/puppetlabs/code/modules --targets localhost #{task_name} #{params}", acceptable_exit_codes: [0, 1]).stdout
+    on(default, "bolt task run --modulepath /etc/puppetlabs/code/environments/production/modules --targets localhost #{task_name} #{params}", acceptable_exit_codes: [0, 1]).stdout
```
ekohl, line 7: Is it still needed to specify the modulepath at all? Isn't this the default built in?

spec/acceptance/suites/scl/redis5_spec.rb
```diff
@@ -4,11 +4,11 @@
 
 describe 'redis', if: %w[centos redhat].include?(os[:family]) && os[:release].to_i == 7 do
   before(:all) do
-    on hosts, puppet_resource('service', 'redis', 'ensure=stopped', 'enable=false')
+    apply_manifest('service{"redis" : ensure => stopped, enable => false}')
```
ekohl, line 7: A closer match (in multi host systems):
```suggestion
    apply_manifest_on(hosts, 'service{"redis" : ensure => stopped, enable => false}')
```

Or keeping it closer to the previous code I think this should still work:
```suggestion
    on hosts, puppet('resource', 'service', 'redis', 'ensure=stopped', 'enable=false')
```

spec/acceptance/redis_cli_task_spec.rb
traylenator: Default was `/root/.puppetlabs` unfortunately.

ekohl: Perhaps works, and let Puppet interpolate it. Not sure how much escaping is needed.
```suggestion
    on(default, "bolt task run --modulepath '/etc/puppetlabs/code/environments/$envirornment/modules' --targets localhost #{task_name} #{params}", acceptable_exit_codes: [0, 1]).stdout
```

traylenator: Tried various level of escaping and brackets with `$environment`

Tried `%x(puppet config print modulepath)` in various ways

PR #476, opened 2023-05-30
Allow deferrables for masterauth & requirepass

### LINKED ISSUE #473: Support deferred values for requirepass
This requires a fairly recent stdlib (8.x), but It would be really useful if the `requirepass` value could be a deferred value.

I'm using this patch in my environment and it works well:

```patch
diff --git a/code/manifests/instance.pp b/code/manifests/instance.pp
index 1f97830..63b5d5f 100644
--- a/code/manifests/instance.pp
+++ b/code/manifests/instance.pp
@@ -327,7 +327,7 @@ define redis::instance (
   Boolean $repl_disable_tcp_nodelay                              = $redis::repl_disable_tcp_nodelay,
   Integer[1] $repl_ping_slave_period                             = $redis::repl_ping_slave_period,
   Integer[1] $repl_timeout                                       = $redis::repl_timeout,
-  Optional[String] $requirepass                                  = $redis::requirepass,
+  Optional[Variant[String, Deferred]] $requirepass               = $redis::requirepass,
   Boolean $save_db_to_disk                                       = $redis::save_db_to_disk,
   Hash $save_db_to_disk_interval                                 = $redis::save_db_to_disk_interval,
   String[1] $service_user                                        = $redis::service_user,
@@ -481,7 +481,7 @@ define redis::instance (
     owner   => $config_owner,
     group   => $config_group,
     mode    => $config_file_mode,
-    content => epp(
+    content => stdlib::deferrable_epp(
       $conf_template,
       {
         daemonize                     => $daemonize,
```

I could send a PR that also adds it in other places (I just patched this one because that's the one I'm using) but honestly I'm not an expert when it comes to Puppet modules, so it would *probably* not contain tests and it would require a fairly big bump of the required stdlib version.

Let me know if this is something you're interested in.

ThiefMaster:
This allows getting the redis password (and/or master password) from an external secret storage using Puppet's Deferred mechanism.

I don't know if there's some way to use `epp()` in case `deferrable_epp()` doesn't exist in order to avoid bumping the stdlib requirement...

closes #473


alexjfisher:
What version of Puppet are you using? From what I've read, this shouldn't be required since Puppet 7.17??

> Remove compiler errors for deferred function mismatched types
Before, it was not possible to compile a catalog that used a Deferred value for a typed parameter class. The compiler would give an error message stating that the type expected did not match Deferred. Now, the compiler inspects the Deferred class's return type and ensures it matches the class parameter type. If the Deferred function has no return type, the compiler warns that it cannot guarantee whether the type adheres to the type the class specifies

See https://tickets.puppetlabs.com/browse/PUP-11518

But... I've not tried yet and perhaps there are still some issues.

traylenator:
Using 7.12 at this environment - it is indeed time we upgraded.

ThiefMaster:
In any case, even without the type requirement failing, I think the `deferrable_epp()` is required for it to work.

PR #492, opened 2023-10-30
Add Debian 12 support

bastelfreak:


evgeni (approved):

ekohl:
Fails because there is no bolt package available. Technically no Puppet package is even available and the distro doesn't package it. I'd be in favor of skipping bolt.

ekohl:
Should we generalize the bolt exception here:
https://github.com/voxpupuli/puppet-redis/blob/364fb02b4d1f846eedb81639dafa1ef2a643772a/spec/spec_helper_acceptance.rb#L10-L14

traylenator:
Obsoleted by #520

evgeni:
520 merged

PR #495, opened 2023-11-09
Rdb save incremental fsync

### LINKED ISSUE #494: `rdb_save_incremental_fsync` cannot be turned off

- Puppet: 7.26.0
- Ruby: ruby 2.7.8p225 (2023-03-30 revision 1f4d455848) [x86_64-linux]
- Distribution: Ubuntu 22.04.3
- Module version: 9.2.0 and `master` 

1. Include the `redis` module:
`include redis` 

2. Add required heiradata to disable `rdb_save_incremental_fsync`
```
redis::rdb_save_incremental_fsync: false
```

## What are you seeing

When `rdb_save_incremental_fsync` is set to `false`, redis should not use `rdb-save-incremental-fsync`. The CLI shows that it is still enabled:
```
$ redis-cli
127.0.0.1:6379> config get rdb-save-incremental-fsync
1) "rdb-save-incremental-fsync"
2) "yes"
```
The relevant config also does not appear in `redis.conf`:

```
# When redis saves RDB file, if the following option is enabled
# the file will be fsync-ed every 32 MB of data generated. This is useful
# in order to commit the file to the disk more incrementally and avoid
# big latency spikes.
<blank line where i would expect to see "rdb-save-incremental-fsync no">
```
the line:
```
rdb-save-incremental-fsync no
```
to appear in `redis.conf` when `rdb-save-incremental-fsync` is `false` or `undef`

the line
 ```
rdb-save-incremental-fsync yes
```
to appear in `redis.conf` when `rdb-save-incremental-fsync` is `true`

## Output log
n/a

I believe this to be happening due to the assumption that `rdb-save-incremental-fsync` is off by default. I have a PR to fix this that'll be along momentarily.

### PR DESCRIPTION
**Pull request by thejambavan**
minor change to the `redis.conf` EPP template to make it possible to disable `rdb_save_incremental_fsync`

Fixes #494

### DIFF
--- a/templates/redis.conf.epp
+++ b/templates/redis.conf.epp
@@ -1030,7 +1030,7 @@ aof-rewrite-incremental-fsync <%= bool2str($aof_rewrite_incremental_fsync, 'yes'
 # the file will be fsync-ed every 32 MB of data generated. This is useful
 # in order to commit the file to the disk more incrementally and avoid
 # big latency spikes.
-<% if $rdb_save_incremental_fsync { -%>rdb-save-incremental-fsync <%= $rdb_save_incremental_fsync ? {true => 'yes', default => 'no'} %><% } -%>
+rdb-save-incremental-fsync <%= bool2str($rdb_save_incremental_fsync, 'yes', 'no') -%>
 
 # Redis Cluster Settings
 <% if $cluster_enabled { -%>

### CODE REVIEW

**Review by kenyon — COMMENTED** [review]
I don't see how this changes anything. If you are providing a non-`undef` value to `redis::rdb_save_incremental_fsync`, then it will render either `rdb-save-incremental-fsync yes` or `rdb-save-incremental-fsync no`.

To be sure, you could add a test like https://github.com/voxpupuli/puppet-redis/blob/e8d71af6e0ae89e53c4ba78456e0c63c95b9d77f/spec/classes/redis_spec.rb#L1683-L1691 that sets `rdb_save_incremental_fsync` to `false` and checks that the config files contains `rdb-save-incremental-fsync no`.

**Review by kenyon — CHANGES_REQUESTED** [review]
Actually I see the problem now. The `if $rdb_save_incremental_fsync` evaluates to false when you set `rdb_save_incremental_fsync` to false, so the line doesn't get rendered. But your code doesn't fix that bug.

**Comment by thejambavan** [comment]
I have corrected the `if` to an `unless Undef` and have added tests which seem to do the job (apologies for multiple formatting commits) for all three states:

Undef: no config file entry, aka "the default" which may have been "no" in the past but is now "yes"
true: yes
false: no

I have separately verified that this works as intended on a real-world instance

File: spec/classes/redis_spec.rb [hunk 1]
```diff
@@ -1690,6 +1690,26 @@ class { 'redis':
...
+          }
+        end
+
+        it { is_expected.to contain_file(config_file_orig).with('content' => %r{^rdb-save-incremental-fsync no$}) }
+      end
+
+      describe 'test rdb-save-incremental-fsync Undef for redis6' do
+        let(:params) do
+          {
+            rdb_save_incremental_fsync: Undef,
```
kenyon, line 1706:
```suggestion
            rdb_save_incremental_fsync: nil,
```

thejambavan:
no matter what i do with the test, when i set:
```
            rdb_save_incremental_fsync: nil,
```
as part of the rspec test, i can't get the test to pass in the Puppet/ruby stages. I don't see any other tests for `undef`/`nil`, and have not had any luck finding good examples elsewhere.

PR #502, opened 2024-02-06
Implement rudimentary Redis and Sentinel ACLs

### LINKED ISSUE #446: ACL support is missing
Redis 6.2 provides way to add other users using aclfile. Will it be possible to provide support for that?

TwizzyDizzy:
This pull request implements the possibility to pass ACLs to redis::instance and redis::sentinel

It does not implement ACL verification / syntax checking, for the overhead of that would IMO not justify the amount of work you would need to put into that.

Closes #446

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -1375,10 +1375,10 @@ Default value: `undef`
 
 Data type: `Array[String[1]]`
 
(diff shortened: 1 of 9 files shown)


rwaffen (approved):
LGTM to me. ubuntu 18 is failing, but this is eol and we should remove this in another pr

PR #503, opened 2024-02-16
remove legacy os'es and add current ones

rwaffen:
remove:
debian10, ubuntu18, rhel7, centos7

add: ubuntu22
(debian 12 might work, but is not testable because we have no official puppet packages)

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -35,21 +35,20 @@
     {
       "operatingsystem": "Debian",
       "operatingsystemrelease": [
-        "10",
-        "11"
+        "11",
+        "12"
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


metadata.json
```diff
@@ -35,29 +35,26 @@
...
-        "10",
         "11"
       ]
     },
     {
       "operatingsystem": "Ubuntu",
       "operatingsystemrelease": [
-        "18.04",
-        "20.04"
+        "20.04",
```
bastelfreak, line 44: please split the changes into two PRs, one that's a breaking change and one that's an enhancement.

rwaffen: okay then :(

PR #504, opened 2024-02-16
remove legacy Debian 10 / Ubuntu 18.04 / CentOS 7

rwaffen:


zilchms (approved):

bastelfreak (approved):

ekohl (commented):
Late to the party, but if you drop these in a major version then please also clean up teh code. For example, dropping EL 7 means you can also drop all the SCL stuff, including parameters.

https://github.com/voxpupuli/puppet-redis/blob/46418870422f4357983ac73e95ca00711736dd5d/manifests/preinstall.pp#L6-L14

Edit: https://github.com/voxpupuli/puppet-redis/pull/509

PR #515, opened 2024-04-16
Support rich datatypes and deferred password values

### LINKED ISSUE #513: Use of deferred function fails due to complex types in epp

- Puppet: 7.24.0
- Module version:

Needs to be run on a puppet server and distinct agent setup, i.e not rspec or puppet apply.

```puppet
redis::instance{String($port):
  service_enable => true,
  service_ensure => 'running',
  requirepass    => Deferred('secret::get',['redis_password']),
  port           => $port,                                      
  bind           => [], # all interfaces
}
```

## What are you seeing

It compiles fine but during the stdlib::deferable_epp part  in the apply

```
Error: Failed to apply catalog: Evaluation Error: Resource type not found: Stdlib::Absolutepath (file: inlined-epp-text, line: 3, column: 3)

```

Should be possible to set password as a deferred to be late evaluated on the agent however it fails due to lack of rich
data types being available on the agent nodes.

The only way I know to "fix" this is to strip back the data types in the template to core data types e..g

`Stdlib::Absolutepath` -> `String[1]`

would be very happy to hear if other options are available.

traylenator:

Process the configuration file `redis.conf` template twice if some of the templates values are deferred.

Currently with deferred values the resulting deferred template cannot be processed since it contains
complex datatypes from stdlib in particular.

This is a redis specific solution that may arrive generally one day in https://github.com/puppetlabs/puppetlabs-stdlib/pull/1425

Fixes #513


manifests/instance.pp
```diff
@@ -501,118 +501,139 @@
...
+    rdb_save_incremental_fsync    => $rdb_save_incremental_fsync,
+    acls                          => $acls,
+    custom_options                => $custom_options,
+  }
+
+  # Rely on https://github.com/puppetlabs/puppetlabs-stdlib/pull/1425
+  # once available.
+  if $_template_params.stdlib::nested_values.any |$value| { $value.is_a(Deferred) } {
+    $_template_params_escaped = $_template_params.map | $_var , $_value | {
+      if $_value.is_a(Deferred) {
```
ekohl, line 613: This is a bit inconsistent. `nested_values` does a deep collection, also going into hashes while the inner part only does the values of the map.

ekohl (commented):
It would be nice to have some sort of test for this. Looking at the rspec-puppet source it should support `Deferred`.

And taking a step back, this feels like `Sensitive` where EPP itself should take care of it.

traylenator:
Tests fail for unrelated reason -  resolved in #516

traylenator:
> It would be nice to have some sort of test for this. Looking at the rspec-puppet source it should support `Deferred`.
> 
Could not see anyway to do something along the lines of:

```ruby
describe 'with parameter requirepass resolved from a Deferred' do
  let(:params) do
    {
      requirepass: Deferred('sprintf',['%s', '_VALUE_']),
    }
  end

  it {
    is_expected.to contain_file(config_file_orig).with(
      'content' => sensitive(%r{requirepass.*_VALUE_})
    )
  }
end

```

> And taking a step back, this feels like `Sensitive` where EPP itself should take care of it.

Not sure I follow. We already test that if requirepass is Sensitive that the content of resulting config file is also marked Sensitive.

ekohl:
> Not sure I follow. We already test that if requirepass is Sensitive that the content of resulting config file is also marked Sensitive.

I was suggesting that Puppet itself should take care of this, just like it can unwrap sensitive. Pushing that responsibility to modules is likely to cause problem somewhere.

Of course it doesn't hold this PR back, maybe more ranting.

ekohl (approved):

traylenator:
CentOS 9 is a timeout.

PR #520, opened 2024-04-28
Add Debian 12 Support

traylenator:

Currently bolt is unreleased for Debian 12.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -35,7 +35,8 @@
     {
       "operatingsystem": "Debian",
       "operatingsystemrelease": [
-        "11"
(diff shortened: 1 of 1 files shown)


ekohl (commented):
I'd prefer some helper like `bolt_supported` that encapsulates that bolt packages are available

evgeni:
> I'd prefer some helper like `bolt_supported` that encapsulates that bolt packages are available

This happened in [#521](https://github.com/voxpupuli/puppet-redis/pull/521)

evgeni (approved):

PR #528, opened 2024-05-15
(#527) Add masteruser parameter

### LINKED ISSUE #527: Not able to specify masteruser parameter

- Puppet: 7.2+
- Ruby:
- Distribution: RedHat 9
- Module version: 11.0.0

Can't find parameter with name masteruser when searching the repo.

## What are you seeing
Zero matches

As of Redis 6 when ACL was introduced and sentinel got support for the following parameters:

Redis:
masterauth <master-password>
masteruser <username>

Sentinel
sentinel auth-user <master-name> <username>
sentinel auth-pass <master-name> <password>

So I expect to be able to specify this for sentinel and redis replicas and so on.

mrsteffenjo:
Enable setting the masteruser parameter which was introduced in Redis 6+ to be able to connect using the new ACL rules.

Fixes #527


mrsteffenjo:
I don't understand why the target configuration holds the following:
```
        +# refuse the slave request.
        +#
        +# masterauth <master-password>
        +
        +# However this is not enough if you are using Redis ACLs (for Redis version
        +# 6 or greater), and the default user is not capable of running the PSYNC
        +# command and/or other commands needed for replication. In this case it's
        +# better to configure a special user to use with replication, and specify the
        +# username configuration as such:
        +#
        +# masteruser <master-username>
        +masteruser _VALUE_
        +# When username is specified, the replica will authenticate against its
        +# master using the new AUTH form: AUTH <username> <password>.
        +
```
that the following test code won't be satisfied:
```
      describe 'with parameter masteruser' do
        let(:params) do
          {
            masteruser: '_VALUE_'
          }
        end

        it {
          is_expected.to contain_file(config_file_orig).with(
            'content' => %r{masteruser.*'_VALUE_'}
          )
        }
      end
```

Oh, I see, my IDE surrounded _VALUE_ with single quote marks.

mrsteffenjo:
Fixed the failed tests

TheMeier:
@mrsteffenjo Thank you for contributing. Sadly that has quite a while has conflicts now. Can you pick this up again?

PR #545, opened 2025-05-15
Allow separate group owner for log directory

sagepe:

Currently the group for `log_dir` is set to the `service_group`. It can be useful to set a different group, eg the `adm` group is often used to allow read access to logs. Setting `log_dir_group` with a suitable `log_dir_mode` would facilitate this.


sagepe:
I just rebased this again - I don't think the test failure is related to the change.

Issue #321, opened 2019-08-15
redis disable snapshots


rmalenko:
1. Disable AOF by setting the appendonly configuration directive to no (it is the default value)
2. Disable RDB snapshotting by disabling (commenting out) all of the save configuration directives (there are 3 that are defined by default)

I have changed for Redis 5.5 this template [redis.conf.erb](https://github.com/voxpupuli/puppet-redis/blob/master/templates/redis.conf.erb) to have uncommented this string `save ""`
```
# Save the DB on disk:
# ...
#   It is also possible to remove all the previously configured save
#   points by adding a save directive with a single empty string argument
#   like in the following example:
#
#   save ""

<% if @save_db_to_disk %>
    <%- @save_db_to_disk_interval.sort_by{|k,v|k}.each do |seconds, key_change| -%>
    save <%= seconds -%> <%= key_change -%> <%= "\n" -%>
<%- end -%>
    <% else %>save ""
<% end %>
```

jeandoBet:
Hi,

Any update on this ? Will it be implemented ?

Thanks,

cjacobsatnwea:
There is a lot more options/etc in redis than this module provides. I have worked around this in my local work by:
* setting up my own redis.conf.puppet-ish file (not the same one by this module)
* inserting exec between `Exec['cp -p /etc/redis/redis.conf.puppet /etc/redis/redis.conf']` and `Service['redis']` where above file is appended to redis.conf

example of the insertion logic (however you add to the redis.conf is up to reader):
```
        notify      => Service['redis'],
        refreshonly => true,
        subscribe   => Exec['cp -p /etc/redis/redis.conf.puppet /etc/redis/redis.conf'],
```
paths might be different for your configs, and above could use vars defining those of course.

ThiefMaster:
I'll use my own template. But seriously, this issue is open for nearly 4 years now. Can't something as trivial as this be impelemented?!

kenyon:
> I'll use my own template. But seriously, this issue is open for nearly 4 years now. Can't something as trivial as this be implemented?!

Not if nobody creates a pull request.

bwitt:
just wanted to clarify, what exactly is needed here?

PR #566, opened 2025-11-05
Parameterize `appendonly` and `save_db_to_disk`

bwitt:

Allow for setting appendonly and RDB snapshotting by making them params.

FIxes https://github.com/voxpupuli/puppet-redis/issues/321


bwitt:
I was trying to allow for disabling redis append only file and snapshotting like it was mentioned in the issue; is this not needed?

kenyon:
If the values are the same for every OS in `params.pp`, then the parameters don't need to be in `params.pp`.

Adding tests is good though.

PR #567, opened 2025-12-19
Allow to adjust the OOM score calculation for redis

dabelenda:
Allow to set the OOMScoreAdjust for the systemd unit.

This allows to control how liklely oom-killer will target redis when in low memory situation.

N/A

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -190,6 +190,8 @@
 #   Specify the time after which a service startup should be considered as failed.
 # @param service_timeout_stop
 #   Specify the time after which a service stop should be considered as failed.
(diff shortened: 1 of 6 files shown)


manifests/init.pp
```diff
@@ -438,6 +440,7 @@
   String[1] $service_user                                        = 'redis',
   Optional[Integer[0]] $service_timeout_start                    = undef,
   Optional[Integer[0]] $service_timeout_stop                     = undef,
+  Optional[Integer] $service_oom_score_adjust                    = undef,
```
bastelfreak: are negative and positive integers allowed? Otherwise we could enforce this via the datatype.

PR #579, opened 2026-02-05
Make it possible to install sentinel independently

mouchymouchy:
Hi there

This PR is a retake of https://github.com/voxpupuli/puppet-redis/pull/389

--------

This PR introduces the possibility to install redis::sentinel standalone, without redis::server. It comes in handy when you need your sentinels to be for example on a different node than your redis server.

The change itself is a pretty straightforward, I am just adding a contain_redis parameter, that makes the requirement of redis class conditional. The default value is true, which includes the redis and it is therefore backward compatible with the current setup.

The only downside of this approach I see currently is that in case someone needs a standalone redis-sentinel from a managed repository, the repository must be added manually as the redis::preinstall is no longer included (as was previously with the redis class)
This Pull Request (PR) fixes the following issues

    installation of standalone sentinel (no open issue for this one, so far)

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,15 @@ class { 'redis::sentinel':
 }
 ```
 
+If installation without redis-server is desired, set `manage_redis` parameter to false, i.e
+```puppet
+class { 'redis::sentinel':
+  ...
+  manage_redis => false,
+  ...
+}
+```
+
 ### Soft dependency
 
 When managing the repo, it needs [puppetlabs/apt](https://forge.puppet.com/puppetlabs/apt).
--- a/manifests/sentinel.pp
+++ b/manifests/sentinel.pp
@@ -150,6 +150,10 @@
 #
 #     user USERNAME [additional ACL options]
 #
+# @param manage_redis
+#   Manage redis base class. If set to false, sentinel is installed without redis server.
+#
+#
 # @example Basic inclusion
 #   include redis::sentinel
 #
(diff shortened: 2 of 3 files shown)


manifests/sentinel.pp
```diff
@@ -205,22 +209,27 @@
   Optional[Variant[String[1], Sensitive[String[1]]]] $sentinel_auth_pass = undef,
   Optional[String[1]] $sentinel_auth_user = undef,
   Array[String[1]] $acls = [],
+  Boolean $contain_redis = true,
```
yakatz: To match other modules, this should be called `manage_redis`. (Users don't know what `contain` means.)

mouchymouchy:
Hi there
Can I have a review please ? :)

README.md
```diff
@@ -133,6 +133,15 @@ class { 'redis::sentinel':
...
 
+If installation without redis-server is desired, set `contain_redis` parameter to false, i.e
+```puppet
+class { 'redis::sentinel':
+  ...
+  contain_redis => false,
+  ...
+}
+```
+
```
kenyon:
```suggestion
If installation without redis-server is desired, set `manage_redis` parameter to false:

```puppet
class { 'redis::sentinel':
  ...
  manage_redis => false,
  ...
}
```

```

mouchymouchy: Ohh good catch you're right :) 
Corrected !

yakatz (approved):

PR #584, opened 2026-04-12
Don't manage ulimit with systemd on FreeBSD

rvstaveren:
Set ulimit_managed class parameter to redis::params::ulimit_managed and set appropriate default per platform in params.pp
This avoids any systemd resource being triggered thus allowing the module to work again on FreeBSD

FWIW, I've ran the lint/tests on Ubuntu 24.04 (because of the Ruby 3.2 requirements during `bundle install` as per https://voxpupuli.org/docs/how_to_run_tests/ ) and observed no issues

```
$ bundle exec rake lint
$ bundle exec rake rubocop
...
24 files inspected, no offenses detected
Finished in 1.43773 seconds
$ bundle exec rake test
...
Finished in 10 minutes 50 seconds (files took 1.92 seconds to load)
2360 examples, 0 failures


2666 examples, 0 failures

```

No issue was filed for this PR, direct fix. However it is somewhat a continuation of #455 with the bare minimum of work to get it going again on FreeBSD

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -2147,6 +2147,7 @@ The following parameters are available in the `redis::instance` defined type:
 * [`output_buffer_limit_slave`](#-redis--instance--output_buffer_limit_slave)
 * [`output_buffer_limit_pubsub`](#-redis--instance--output_buffer_limit_pubsub)
 * [`custom_options`](#-redis--instance--custom_options)
+* [`config_dir`](#-redis--instance--config_dir)
 
 ##### <a name="-redis--instance--activerehashing"></a>`activerehashing`
 
@@ -3182,6 +3183,14 @@ Data type: `Hash[String[1],Variant[String[1], Integer]]`
 hash of custom options, not available as direct parameter.
 
 Default value: `{}`
+
+##### <a name="-redis--instance--config_dir"></a>`config_dir`
+
+Data type: `Stdlib::Absolutepath`
+
+
+
+Default value: `$redis::config_dir`
 
 ## Functions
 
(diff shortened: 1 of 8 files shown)


bastelfreak:
Hi, thanks for the PR. Please run `bundle exec rake strings:generate:reference` to update the REFERENCE.md

bastelfreak:
Can you please add a unit test and add Freebsd to metadata.json?

rvstaveren:
FYI: Put to draft while I iron out the defects presented by the defects

rvstaveren:
Getting the tests right was quite a rabbit hole. The ubunti2404 test seems to fail in this pipeline, but I was able to run it locally as `env BEAKER_SETFILE="ubuntu2404-64{hostname=ubuntu2404-64-openvox8.example.com}" bundle exec rake beaker` resulting in a 

```
Finished in 6 minutes 59 seconds (files took 17 minutes 53 seconds to load)
50 examples, 0 failures
```
Not sure what is going on here, but there might be some discussion needed on [aff8299](https://github.com/voxpupuli/puppet-redis/pull/584/commits/aff82996cb63fb2ec5313ad8f97e896b67c7d48e) and related [914a999](https://github.com/voxpupuli/puppet-redis/pull/584/commits/914a9998b8c798459a6f1c2713ef53fd3df32378) in where RHEL8 will start to use /etc/redis (or valkey) for their instance configuration. I made a rocky8 vm and saw that /etc/redis is created as part of the module invocation. 

Not having `dirname` around any longer in manifests/instance.pp simplifies the tests and makes it possible for the FreeBSD adjustments to land there, keeping it and trying to add an case for FreeBSD will raise a spec alert for a “too many local variables”. It is nevertheless a change for the RHEL8 systems and might need a PR of its own instead of piggybacking on the changes for FreeBSD. 

So we can either
* let the `dirname` change piggyback in this PR
* Do the `dirname` change in a separate PR and use that as the foundation for this PR, which will be kept on draft

What do you think?

Issue #575, opened 2026-01-10
Unclear that valkey is supported?


bwitt:
It's perhaps a bit unclear that valkey is supported by this module.  Neither the README nor REFERENCE.md mentions it.

Maybe we should put a note in one or the other?

Ref https://github.com/voxpupuli/puppet-redis/issues/573

PR #587, opened 2026-05-27
Migrate params to hiera layers with extended provider switch support

cruelsmith:
Remove `params.pp` and migrate everything to hiera data layers.

Using lookup() and alias() function inside hiera for extented provider switch support.

Notes:
- RedHat - EL7 support will be broken after this since default provider in `hiera/os/RedHat.yaml` is `valkey` now.
- Real world testing if just the provider change switch does work on any OS has not be done. We just assume here that every `redis` naming has just been replaced with `valkey`.

None

Relates to #575


kenyon (changes requested):
The values in `common.yaml` should instead be the class parameter defaults in `init.pp`. We do this so that the defaults are documented in `REFERENCE.md`.

cruelsmith:
> We do this so that the defaults are documented in `REFERENCE.md`.

Historic yes but i was also stunned to see that lastest tool versions for generating `REFERENCE.md` now also includes the default values from `comman.yaml` :sweat_smile: 

See https://github.com/voxpupuli/puppet-redis/pull/587/changes#diff-6269a4217141bc6fd747b87032c8f248c977cfd8b2f0d0d7d1ba04af62ba229aR332

https://github.com/voxpupuli/openvox-strings/releases/tag/v7.0.0

----

So should i still move them to `init.pp` or do we keep them like this?
In cases like this
```
redis::sentinel::service_group: '%{alias("redis::service_group")}'
```
it would not require to pull in the `init.pp` as dependency in the `sentinel.pp`. Currently it is optional to be added there.

kenyon:
Ah I didn't know about that improvement to openvox-strings. There is a question about the style here: https://github.com/OpenVoxProject/openvox-docs/issues/216

I guess I am fine with it either way, with a slight preference for the defaults in the class parameters so that they are all in one place.

kenyon (approved):

